PR #25, opened 2015-03-23
Mention David Kastrup in "Promoting Git developers"

moy:
My feeling reading the list is that the discussion leading to this rev news was actually initiated by David Kastrup's message, hence he should be the one to credit here (even though his message insisted more on the problem than on solutions).


tfnico:
LGTM. @chriscool merge at your discretion.

chriscool:
LGTM too and merged, thanks all!

PR #27, opened 2015-03-23
revnews: talk about &&-chain test

gitster:


tfnico:
Great piece. I'm sort of opinion-less on the whole attribution thing, so LGTM. @chriscool?

gitster:
It is perfectly OK to drop the tip commit and merge the main text only. I just wanted to illustrate what I meant with an example.

If we were to do this "per-article attribution" thing, you may want to give attribution to all "outside" contributions (where the definition of "inside" being "the editorial team"). That way, people may be enticed to write for future editions, hopefully lowering the load on the editorial team.

chriscool:
@gister yeah I agree with your reasoning. Thanks!

chriscool:
I also commited this:

https://github.com/git/git.github.io/commit/765180653a2d21d703e3340f905c582bfbf459ac

PR #28, opened 2015-03-24
revnews: mention some recent releases

tfnico:
I figure it is interesting to mention recent Git "core" releases, 
as well as in related projects (libgit2, jgit, etc).

PR for @chriscool, since it is a larger block of text.

### DIFF
--- a/rev_news/draft/edition-1.md
+++ b/rev_news/draft/edition-1.md
@@ -267,6 +267,34 @@ to GitHub and the sponsors. If you are a Git developer and need travel
 or other assistance to go there, please contact Peff, aka
 Jeff King &lt;<peff@peff.net>&gt;.
 
+### Releases
+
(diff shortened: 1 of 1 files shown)


chriscool:
@tfnico ok, but then "Commandline GUI-tool tig version 1.2 was released." could be moved in the release section too?

chriscool:
By the way, maybe we could use "Media" or "Other" instead of "From outside the list". I think it would be simpler and more inline with the other section titles.

tfnico:
@chriscool Sorry for not responding earlier. +1 on all points. Thanks.

chriscool:
Ok, thanks for responding.

As you can see, I added a patch to move the tig news into the Releases to this PR and I merged it before publishing edition 1.

About using "Media" or "Other" instead of "From outside the list", I will let's say use "Other" in edition 2.

PR #40, opened 2015-04-10
Prettier design

tfnico:
I've grabbed some CSS from the libgit2 page, basically the same setup that they have on the git-scm.com page as well, and worked it into our site.

The layout is pretty much the same, but the colors and fonts are prettier, at least.

I also changed all the quotes in edition 1 to be blockquotes, because that's what they are. I did make them monospaced, so stuff from the mailing list still looks like monospaced mailing list text.

Please note that I hardly know anything about CSS, and now I know even more than I ever wanted to know. Or something.

@chriscool Please take a look.

### DIFF
--- a/_includes/README.md
+++ b/_includes/README.md
@@ -19,11 +19,6 @@ git@vger.kernel.org if appropriate).
   `/usr/bin/ruby` and `/usr/bin/gem` with their v2 options)
 * `sudo gem install bundler`
 * Clone this repo
-* If I haven't committed the Gemfile yet, create it with these lines:
-```
-source 'https://rubygems.org'
-gem 'github-pages'
-```
 * `sudo apt-get install zlib1g-dev` # ref [1]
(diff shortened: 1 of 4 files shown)


chriscool:
Ok, thanks a lot again!

chriscool:
I must say that overall it looks much better, but I am not a big fan of having some headlines written in red. What do you think?

tfnico:
Sure, I'll see if I can make a different color.

chriscool:
Great, maybe dark blue or just black?
Thanks anyway!

tfnico:
@chriscool Changed it to black now, and tweaked the h2/h3 font styles so they are easier to tell apart. Ref https://github.com/git/git.github.io/commit/d4c273045a6a68e949d19ea35458f42b1384d5fc

chriscool:
Yeah! I like it very much now!
Thanks!

PR #41, opened 2015-04-12
Tiny fixes to wording of General section

durdn:
I did a quick pass rewording a couple of paragraphs. Feel free to discard if you want to maintain a single consistent voice.

### DIFF
--- a/rev_news/draft/edition-2.md
+++ b/rev_news/draft/edition-2.md
@@ -11,9 +11,9 @@ in a format that the wider tech community can follow
 and understand. In addition, we'll link to some of the interesting Git-related
 articles, tools and projects we come across.
 
-This special edition covers Git's 10th anniversary and
-the [Git Merge 2015](git-merge.com) on April 8th & 9th in Paris, France,
-where many Git fans celebrated this anniversary. 
+This special edition covers Git's 10th anniversary and the [Git Merge
+2015](git-merge.com) held on April 8th & 9th in Paris, France, where many Git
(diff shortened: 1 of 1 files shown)


durdn:
@gitster sorry about that, indeed my bad vim `gq` re-wrapping habit took over. I'll split the two actions next time.

tfnico:
Out of curiosity, what is the general solution for that problem? Should one first check in bad line lengths while fixing wording, and thereafter correct line wrapping in a second commit?

gitster:
@tfnico Yeah, if you really want to rewrap, then do the real fix as-is and then rewrap in a separate commit would be the way to go, as doing these in opposite order would not make much sense (the real fix likely will make the line length uneven again).

But we are dealing with markdown whose final rendering will wrap lines appropriately, so
not doing any rewrapping would be the way to go, IMNSHO ;-)

peff:
Actually, markdown wrapping is not consistent from flavor to flavor. Some flavors will keep hard newlines inside a paragraph (and expect each paragraph to be on its own line). This is helpful for clueless folks pasting text into a browser input box, but I find it intensely irritating.

I noticed this when test-rendering #54 (i.e., the "view" button on GitHub), but it looks like there is hard-wrapping throughout the document as rendered by Jekyll (find the "Paraphrasing Junio" line, for example). You can replace the markdown renderer in Jekyll; I wouldn't be surprised if you can also tweak the wrapping behavior.

PR #47, opened 2015-04-13
edition-2: fix rendering of underscore

dscho:
The underscore means *italic* in Markdown, and the GitHub Pages renderer interprets underscores even in the middle of terms as such. Let's fix the funny rendering by escaping the underscores.

### DIFF
--- a/rev_news/draft/edition-2.md
+++ b/rev_news/draft/edition-2.md
@@ -121,11 +121,11 @@ becomes much faster to have as much information as possible in a
 single file. That's the purpose of the packed-ref file.
 
 Peff discovered that one of his own commit that switched from fgets()
-to strbuf_getwholeline() to read the packed-ref file was in part
+to strbuf\_getwholeline() to read the packed-ref file was in part
 responsible for a big slowdown.
 
-strbuf_getwholeline() is part of the Git strbuf API that is used for a
-lot of string related functions. And strbuf_getwholeline() used the
+strbuf\_getwholeline() is part of the Git strbuf API that is used for a
+lot of string related functions. And strbuf\_getwholeline() used the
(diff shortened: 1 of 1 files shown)


tfnico:
Well spotted! Thanks.

peff:
I don't think it's worth changing here, but for future reference, backticks in markdown introduce a "code" span where punctuation is interpreted literally. So:

    This is a paragraph about `strbuf_getline()`, whose `_` does not
    match the one in this `strbuf_getline()`.

renders as:

This is a paragraph about `strbuf_getline()`, whose `_` does not match the one in this `strbuf_getline()`.

That's assuming you like the extra typesetting given to code spans (which I do).

dscho:
> I don't think it's worth changing here, but for future reference, backticks in markdown introduce a "code" span where punctuation is interpreted literally

Good point. I actually *had* considered doing that, but for consistency, I would have had to edit the entire document to use backticks for *all* of the code references and I was too lazy ;-)

tfnico:
@dscho @peff @chriscool The reason this wasn't done using backticks in the first place was that we used "pre" (4 space indentation) originally for quotation. This weekend, along with the new design, I replaced all quotations with blockquotes, and thereby the preformatted text went all bonkers.

We'll make sure to format properly in the future.

dscho:
@tfnico hey, I am happy I could contribute something! You already described the Git for Windows release candidate, so I had nothing else left than this fix. ;-)

peff:
@tfnico Ah, that makes sense. I think the blockquote looks nicer, but it does mean that importing text from people's messages will require manual markdown cleanup (it _usually_ does what the original author intended, but of course not always :) ). I guess that sort of cleanup is not too hard, though, as the content is being proofread anyway.

PR #48, opened 2015-04-13
Remove violent language; replace it with the desired action.

emmajane:
It is not necessary to ask people to beat software, when the desired outcome is actually
to have people test the software. Using clear, positive language will help reinforce the
friendly, helpful behaviour I experienced at Git Merge.

### DIFF
--- a/rev_news/draft/edition-2.md
+++ b/rev_news/draft/edition-2.md
@@ -190,8 +190,9 @@ working on making Git fast for big repositories.
 * [First release candidate for Git 2.4](http://git-blame.blogspot.de/2015/04/first-release-candidate-for-git-24.html),
   April 2nd.
 
-> Please give it a good beating so that we can ship a successful v2.4 final at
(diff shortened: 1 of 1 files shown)


tfnico:
Thanks Emma!

emmajane:
Aha! I had misunderstood that it was a direct quote from the link. I agree that it should either be a direct quote, attributed; or a directive from the newsletter which uses "plain language". If I were sugarcoating it would look much different again. ;) I'm happy to submit a new PR, according to @gitster's preference (as the source of the quote, I think it makes sense to let the author choose).

tfnico:
According to the comment here  https://github.com/git/git.github.io/commit/6269b5b2f9fc13eb529f1faf892868c1b1c8c9c7#commitcomment-10708448 I think everyone's happy now.

Next time I'll merge a bit less fast. But let me underline how I really appreciate Emma's initiative to help us here :)

emmajane:
<3

Onwards! :)

PR #54, opened 2015-04-16
rev-news: fixup quote regarding regressions from Junio

peff:
In paraphrasing a quote in 3bfd3a01054d3448aefdc148c3bb5e7a579935de, we lost the original meaning. There weren't regressions in v2.3; the original was trying to say that we care about making sure we don't have regressions since v2.3. Hopefully this paraphrase makes that more clear.

/cc @gitster @emmajane


tfnico:
Thanks, I'm going to merge this preemptively due to @gitster's comment in https://github.com/git/git.github.io/commit/57b792d8deed175a95c6590fa64d2750ebc83d58#commitcomment-10747322

PR #62, opened 2015-05-11
missing word in "The Git Project"

stefanbeller:


stefanbeller:
Maybe it was on purpose though, as it appears again in "what should The Project do with its money?"

tfnico:
I don't care that strongly about it, but I would leave out the capitalization, and just write "the Git project", "the project", etc.

gitster:
I'd say that phrase should be spelled as "the project".

PR #65, opened 2015-05-18
Remove line containing a reminder to copy information

shubhamchaudhary:
All the concerned information is mentioned in both about and readme section now.


tfnico:
Thanks!

PR #68, opened 2015-05-29
rev-news 4: recent activity on git-multimail

moy:


chriscool:
Thanks! I think I or @tfnico will move it to a subsection though, like "Releases" or maybe create a new subsection.

PR #69, opened 2015-05-30
Rephrase "corrupt patches cannot be applied"

gitster:
It is not like reformatted and reflowed patches are ignored by malice; heavy corruption makes them impossible to apply and hard to comment on.


tfnico:
Thanks!

PR #70, opened 2015-05-30
rn-4: stress that submitGit is now safe to use

gitster:
It is not so important that the original was unsafe; the fact that it was fixed and now is safe
is more important, and encourage people to try it out.


tfnico:
Thanks!

chriscool:
I think @tfnico merged this in 6ba6e8e7593a9d381037aa209b2d8bdba0748952

PR #71, opened 2015-05-30
rn-4: review of the "clean/smudge empty contents" topic

gitster:


tfnico:
There already is a section called on "sha1_file: pass empty buffer to index empty file" written by @chriscool - I'll leave it to him to pick the best texts here.

chriscool:
@gitster I merged your version which was better. Thanks!

PR #96, opened 2015-09-04
first draft of text for GSoC wrap up.

moy:
This should eventually become a blog post on
http://google-opensource.blogspot.fr/.

Let's use Git Rev News both as a way to work together on the draft, and as an alternative publication mechanism: readers of Git Rev News may like the text, and may not read Google's blog.

### DIFF
--- a/rev_news/drafts/edition-7.md
+++ b/rev_news/drafts/edition-7.md
@@ -21,6 +21,119 @@ This edition covers what happened during the month of August 2015.
 ### General
 -->
 
+### Google Summer of code Wrap Up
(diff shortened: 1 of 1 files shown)


tfnico:
:+1: 

We may want to inform that it's a cross-post, when we have the URL.

moy:
> We may want to inform that it's a cross-post, when we have the URL.

Yes. I guess the Git Rev News version will be posted before so we'll add it after the publication.

chriscool:
Merged thanks!

PR #102, opened 2015-09-10
rn-7: fix partly broken link to youtube

zephyr:
The original link is „doubled“ (probably a copy-paste error), so that youtube isn’t able to parse the timing part and starts the video at the beginning. So let us:

- remove the repetition,
- remove the reference to the not used youtu.be link shortener,
- and, as much as we all like the metric system, `21m51s` is more readable as `1311`.


tfnico:
Thanks!

zephyr:
Thanks for your work at this newsletter and your git podcast!

PR #107, opened 2015-10-12
Toot my own horn

dturner-tw:
Add discussion of ref backends

### DIFF
--- a/rev_news/drafts/edition-8.md
+++ b/rev_news/drafts/edition-8.md
@@ -237,6 +237,15 @@ I only found this recently, it's just brilliant. Thank you, whoever it was put t
 
 ## Other News
 
+__Updates__
+
+* [Since we talked about it last month](http://git.github.io/rev_news/2015/09/09/edition-7/) David Turner
+submitted [a new patch series](http://thread.gmane.org/gmane.comp.version-control.git/278757) of his work on
+alternate ref backends. This work is being reviewed by Michael Haggerty, Junio, and others. The ultimate goal 
+is still to allow a lmdb-based backend, which should be faster and have fewer issues on
+case-insensitive filesystems than the traditional files-based backend. This work is sponsored by
+Twitter.
+
 __Various__
 
 * The GitMinutes podcast finally published the last episode from the Git-Merge conference. Here are all five parts in a neat overview:


rev_news/drafts/edition-8.md
```diff
@@ -87,6 +87,11 @@ patches](http://thread.gmane.org/gmane.comp.version-control.git/279348),
 so an interesting way to test patchs will perhaps be available soon
 to Git developers.
 
+David Turner submitted a draft of his work on alternate ref backends. 
+This work is being reviewed by Michael Haggerty, Junio, and others.  The 
+ultimate goal is to allow a lmdb-based backend, which should be faster and
```
moy: It may make sense to say "better than what?" here.

dturner-tw: added "than the traditional files-based backend".

chriscool:
There are the following problems with this PR:

* It should be moved elsewhere as it has nothing to do with adding Travis CI support. For example it could be moved under Other News where there are other very short news.
* It should have a link to the discussions on gmane.
* As there was an article about the same topic in [the previous edition](http://git.github.io/rev_news/2015/09/09/edition-7/), it could link to it and should probably be presented as an update to the previous article with some new information.

In other words I would be ok with something like the following:

```
## Other News

__Updates__

* [Since we talked about it last month](add link to previous article or previous edition) David Turner submitted [a new patch series](add link to the discussion on the list) of his work on alternate ref backends. This work is being reviewed by Michael Haggerty, Junio, and others. The ultimate goal is to allow a lmdb-based backend, which should be faster and have fewer issues on case-insensitive filesystems than the traditional files-based backend. This work is sponsored by Twitter.

__Various__

```

chriscool:
Thanks @dturner-tw!

One last request though: could you squash everything into one commit and force push on your refs-backends branch?

(That will avoid polluting the history with four commits instead of one.)

dturner-tw:
Squashed.  Thanks.

gitster:
The topic is a very interesting addition, but why isn't it part of the regular "Reviews" section, I wonder. After all, isn't the primary point of Git Rev News about reporting and summarizing the review discussion threads to give credits to reviewers?

chriscool:
@gitster Yeah David's update is nice, but I think the "Reviews" section should be for articles that are a bit longer than this one.

And yeah, we want want to give credits to the reviewers, but it's better if we describe a bit more what they are doing than just give their names, otherwise we could just have a section with only all the reviewers names. That's why I think we should encourage articles that are a bit longer and put only those in the "Reviews" section.

dturner-tw:
Since this was for the September update, I decided to skip most of the details, since I think the interesting discussions really started this month.  Happy to adjust if necessary.

PR #109, opened 2015-11-08
Improve wording in "file time issues" section

tfnico:
@chriscool Some of these paragraphs felt a bit clumsy, so I tried to improve them here. You like?


tfnico:
Huh, I didn't really mean for 55a136a to be in this PR, but whatever.

PR #122, opened 2016-02-16
Add 2015 to reasoning of being a veteran org

stefanbeller:


stefanbeller:
I think I messed up the dates. We are applying to GSoC 2016, not 2015.

PR #127, opened 2016-02-19
libgit2 projects

carlosmn:
Update the URL since we now have a more "normal" `master` branch and add the two main projects we'd be looking to mentor.

/cc @ethomson


moy:
Perfect, thanks. And even before the deadline ;-).

ethomson:
:+1: Thanks for doing this, @carlosmn !

PR #128, opened 2016-02-22
add step to check tests results on Travis-CI

larsxschneider:


SoC-2016-Microprojects.md
```diff
@@ -66,6 +66,17 @@ the following steps:
   text is contained in `Documentation/SubmittingPatches`), and to
   signify your agreement by adding a `Signed-off-by` line.
```
mendoza6688: Buenisima

PR #130, opened 2016-03-07
soc-2016-ideas: adjust markdown for "whitelist" example

peff:
GitHub's markdown seems to render this as one big jumble, rather than as a `<pre>`-formatted block.

I'm not really sure _why_, but dropping the indentation seems to fix it.

### DIFF
--- a/SoC-2016-Ideas.md
+++ b/SoC-2016-Ideas.md
@@ -291,18 +291,18 @@ and blacklist using Git config.
 Whitelist example:
 ```
 [remote]
-    default = deny
-    message = "Are you sure you're not pushing company code?"
-    allowed = http://whitelisted-hosting.org
-    allowed = http://git-hosting.org/whitelisted-org
-    allowed = http://git-hosting.org/org/whitelisted-repo
+default = deny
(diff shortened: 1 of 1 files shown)


peff:
Actually, I'm not sure that it does fix it. It looks fine in the github.com markdown preview, but so does the original. I wonder if this is some Jekyll-related weirdness. Is `[foo]` at the start of a code block special or something?

peff:
After some local testing with `jekyll`, I think the right fix is in d6cc5b7 (the fenced blocks need to be separated from the preceding paragraph).

Arguably this is a bug in redcarpet, but it's easy for us to work around (and I think the resulting source is more readable anyway).

Interestingly, I triggered the bug on my local build only by setting `markdown` to `redcarpet` in the config. The default these days is supposed to be `kramdown`, but using `kramdown` locally looked totally broken (it did not handle fenced code blocks at all!).

So fingers crossed that this looks good when I merge and deploy it. :-/

peff:
Still puzzled about the details, but this _did_ fix it on the production site. Yay?

PR #132, opened 2016-03-08
Balance parentheses in Outreach-2016-May page

gitster:


peff:
Eep. I swear I proofread. :)

PR #134, opened 2016-03-12
SoC-2016-Microprojects: explain how to search for microprojects

chriscool:
I think it might be interesting to see if some students can come up with interesting microprojects by themselves given some broad instructions explaining how they could search for some microprojects.


chriscool:
I was inspired by reading this email from @gitster:

http://thread.gmane.org/gmane.comp.version-control.git/286197/focus=286510

and by the way I wonder if we could still add a microproject to add a "--nul" alias to git config's "--null" and deprecate "--null"?

peff:
I think these are all good tips (though looking for `test_expect_failure` may sometimes lead to rabbit holes; if we bothered to document a failure but not fix it, that is often because the fix is non-trivial).

PR #140, opened 2016-03-13
soc-2016-ideas: add URL rewrites project

larsxschneider:


larsxschneider:
rebased, to fix merge conflict.

larsxschneider:
@peff Should I post this suggestion to the mailing list, first?

peff:
It certainly can't hurt. I don't know if people are actively following PRs here, so you will get a lot more feedback from the list.

peff:
I didn't see any mention on the list. But having read it over, I think this is probably a reasonable thing to be doing, but isn't nearly big enough for a GSoC project (and is probably too big for a micro-project).

moy:
I agree with peff here: it's too small for a GSoC. Why not add it to the idea page, but with a disclaimer that it's not big enough and can only be a part of a GSoC (e.g. a project about "git config improvement" could include this as a sub-project). Students don't really get a second chance after they submitted their proposal, so attracting them towards a project that has no chance of getting accepted as-is is a bit nasty.

Also, I think it's too late for this year's proposals: students should already have gone through the list and started to work on a proposal (ideally long ago). But it's nice to start working on next year's right away ;-).

Thanks,

larsxschneider:
@moy: I agree, we shouldn't post something that would not be accepted. Thanks for the hint!
@peff: I was too busy to post it on the list, yet.

Since posting this as GSoC project this year would not make sense anymore... what if we create a dedicated page for "Git project ideas". This page could be an inspiration for next years GSoC/Outreachy as well as for people that are just looking for their next challenge.

larsxschneider:
... or maybe we create GitHub issues and ask people to use "reactions" to vote... this could be a way for non git-list people to express what they think could be an interesting new feature ...

moy:
@larsxschneider : just a ping. It's obviously too late for this year, but it makes sense to me to add ideas so that we have more ideas to start with next year. I think it makes sense to merge something like this, but to add a clear warning about the fact that it's too small for a GSoC and it should be grouped with something else.

PR #146, opened 2016-03-23
Soc instructions

moy:
Too late for this year, but we have so many wrong applications, let's try to clarify them now so that we have a better template next year. What do you think?

(we can merge this right after the deadline)

### DIFF
--- a/SoC-2016-Microprojects.md
+++ b/SoC-2016-Microprojects.md
@@ -19,6 +19,18 @@ libgit2 list of
 projects](https://github.com/libgit2/libgit2/blob/master/PROJECTS.md)
 rather than the list below.*
 
+Git development is based on sending successive versions of patches or
+patch series to the mailing list until they are considered good and
+correct by the reviewers and Junio Hamano, the maintainer, who will
+merge them. This process usually takes quite some time. By sending
+drafts of your microproject patches to the
+mailing list long before the deadline, you can show us that you are
+willing and able to work well using the Git development process.
+
+It is *expected* that what you send will need several rounds of
+reviews and discussions. If you are not sure at all about a patch you can
+put "[RFC/PATCH]" at the beginning of its subject.
+
 Consider [a sample email
 thread](http://thread.gmane.org/gmane.comp.version-control.git/239068),
 which shows how a developer proposed a change and a patch to implement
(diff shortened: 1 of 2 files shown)


peff:
Thanks, I think all of the changes here make sense. :+1:

moy:
The deadline was friday. I added commits by @chriscool. Everybody's OK?

SoC-2016-Microprojects.md
```diff
@@ -19,6 +19,18 @@ libgit2 list of
 projects](https://github.com/libgit2/libgit2/blob/master/PROJECTS.md)
 rather than the list below.*
 
+Git development is based on sending successive versions of patchs or
```
peff: Usually we'd say `patches` in English (I thought initially it was just a typo, but TIL that it's `patchs` in French).

SoC-2016-Microprojects.md
```diff
@@ -19,6 +19,18 @@ libgit2 list of
...
+patch series to the mailing list until they are considered good and
+correct by the reviewers and Junio Hamano, the maintainer, who will
+merge them. This process usually takes quite some time. By sending
+drafts of your patches for your microproject, to the
```
peff: I think this comma is extraneous. Might also be easier to parse as `sending drafts of your microproject patches to the mailing list...` to reduce the recursion level of prepositional phrases.

SoC-2016-Microprojects.md
```diff
@@ -19,6 +19,18 @@ libgit2 list of
...
+mailing list long before the deadline, you can show us that you are
+willing and able to work well using the Git development process.
+
+It is *expected* that what you send will need several rounds of
+reviews and discussions. If you not sure at all about a patch you can
```
peff: `s/you not/you are not/`

peff:
I pointed out a few minor issues, but other than that, feel free to merge.

By the way, you may want to just start pushing to non-master branches directly in the git/git.github.io repo, and opening PRs from there. If you do, then other collaborators on the project can push onto the branch to update the PR, too (and I probably would have done so for minor things like typo fixups).

Of course, if you'd prefer people not push onto your branches, then don't. :)

moy:
Applied your remarks. Indeed, pushing to non-master branch in this repo would be better.

Thanks,

PR #148, opened 2016-03-25
SoC: use 'criticizing' instead of 'critiquing'

chriscool:
It seems to me that 'critiquing' is not correct.

cc @peff @moy

### DIFF
--- a/SoC-2016-Ideas.md
+++ b/SoC-2016-Ideas.md
@@ -17,7 +17,7 @@ applicants become familiar with Git's development and submission
 process.
 
 Also, working in Git project is not only about writing your own
-patches. Constructively critiquing design and implementation of
+patches. Constructively criticizing design and implementation of
 patches by other people is also an important skill you need to learn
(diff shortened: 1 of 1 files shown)


moy:
It seems it is actually correct actually:
http://public.wsu.edu/~brians/errors/critique.html
http://www.wordreference.com/enfr/critique

But I'll let native speakers decide ;-).

chriscool:
Yeah it is strange that there is also criticize: http://www.wordreference.com/enfr/criticize

peff:
I think "criticize" was the original verb, but like many things, we Americans managed to start using the noun form as a verb. These days, though, I think I would prefer "critique"; it implies a constructive objectiveness to the feedback, whereas "criticize" generally implies strict negativity. I have no idea if that is different in British English, though.

chriscool:
Ok, let's just keep "critiquing" then.

fugazyrappersoutlook:
https://github.com/MichaelAquilina/Reddit-Data/pull/3

PR #150, opened 2016-04-18
Fix some small typos and make some grammar changes

Cogito:


rev_news/drafts/edition-14.md
```diff
@@ -117,65 +116,65 @@ He is the maintainer of the Kernel '-stable' branches and of many
...
+Around 10 000 lines are added, 5300 lines are removed, and 1800 lines
 are modified, everyday!
 
 That's on average 7.8 changes per hour accross the whole tree with 5%
-in the core, 10% in the networking subsystem and 55% in the drivers.
+in the core, 10% in the networking subsystem, and 55% in the drivers.
 
-This goes against any previously thought methodology for stable
-software. And things are going faster and faster.
+This goes against all previous methodologies for stable
```
chriscool: About this line, I'd rather keep what I wrote as I think it is very close to what Greg said.

rev_news/drafts/edition-14.md
```diff
@@ -117,65 +116,65 @@ He is the maintainer of the Kernel '-stable' branches and of many
...
-are fixed a release is made and the cyce starts over with a ne merge
+are fixed a release is made and the cycle starts over with a new merge
 window.
 
 For "stable" kernels that Greg maintains, they are forked from Linux
 releases. Commits have to go in Linus' tree first before Greg will
-accept them. It should be the identical patch as what is in Linus'
-tree and it should be bug fixes only or new device ids. The Linux
+accept them. It should be the same identical patch as what is in Linus'
+tree, and it should only contain bug fixes or new device ids. The Linux
```
chriscool: IIRC "bug fixes only" is what Greg said, so I prefer to keep it.

Cogito:
I saw a couple of typos, and so went in to fix them, but ended up doing more general editing. Happy for you to take any parts you think seem good.

rev_news/drafts/edition-14.md
```diff
@@ -117,65 +116,65 @@ He is the maintainer of the Kernel '-stable' branches and of many
...
-smallest possible patches and every change should be correct.
+None of them should break the build. Each patch set should be obvious,
+broken down into the smallest reasonable patches, and every change
+should be correct. One half to one third of submitted patches get accepted. 
 
-This puts work on the developer's shoulders. But that's on purpose
-because there are much more developers than maintainers. There are
-1000 maintainers but only around 700 are active.
+This puts most of the work on the developer's shoulders. That's on purpose,
+as there are many more developers than maintainers. There are
```
chriscool: Ok for s/much/many/ but otherwise I'd rather keep a version that is closer to what Greg said.

chriscool:
Great! Thanks a lot!
I will merge most of this.

Cogito:
I think most of it is is just a matter of tone. That is, the report is in some places a recount of what was said and in others a summary. I think that generally comes across in the tone, but sometimes it meant that there were lots of short sentences and odd phrasing.
This probably kept true to what was said, but might sacrifice readability.
My only suggestion in those cases is to either rephrase those pieces to be a summary, rather than a recount (so it doesn't matter if it sounds similar to the original), or to explicitly quote what was said. I think the phrasing jumped in and out of summarising and recounting at some points, so being explicit in those areas would probably help the reader follow what's happening.

chriscool:
Yeah, it's true that it looks like a summary in some places. That's because I couldn't note or remember everything that was said. And I'd rather not quote a lot of sentences as I think it would also sacrifice some readability, and also I might have not recalled/noted everything properly.

I don't think there is a perfect solution here, as I don't want anyway to record everything and transcript it. And I think it is mostly ok as it is now. We are not into professional journalism anyway.

Cogito:
Yep, I think it reads pretty well in any case, and I agree about not quoting too much.

I'm guessing there is a part 2 (and more?) to come later??

chriscool:
Yeah, I plan to report again on the Contributor Summit and the Git Merge in the following edition (edition 15 next month).
Thanks again!

PR #154, opened 2016-05-09
Small fixes for edition 15

Cogito:
This pull request includes a few grammar changes, typo fixes, and expansion of some examples to make them easier to follow.

### DIFF
--- a/rev_news/drafts/edition-15.md
+++ b/rev_news/drafts/edition-15.md
@@ -91,7 +91,7 @@ Some of the tricks he showed are available in his
 The presentation started with simple shell aliases like `alias
 sl='ls'` to cope with typos and `alias g='git'` to reduce typing, then
 a bit more complex ones to avoid typing common arguments like `alias
-gprom='git push origin master'` and `alias pydoc='python -m pydoc'`,
+gpom='git push origin master'` and `alias pydoc='python -m pydoc'`,
 until validation aliases like:
 
 ```
--- a/rev_news/drafts/edition-15.md
+++ b/rev_news/drafts/edition-15.md
@@ -98,8 +98,9 @@ until validation aliases like:
 alias yamlcheck='python -c "import sys, yaml as y; y.safe_load(open(sys.argv[1]))"'
 ```
 
-Then Spencer spoke about customized prompts for git, filesystems and
-tools, and about git config aliases, for example:
+Spencer then spoke about customized prompts for git, filesystems and
+tools, and about git config aliases. For example, this alias that makes
+grepping through your submodules easy:
 
 ```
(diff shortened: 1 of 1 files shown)


rev_news/drafts/edition-15.md
```diff
@@ -27,52 +27,52 @@ Git Merge conference on April 5 2016.
...
-their obligations. In case some developers who contributed to Git
-wants to take a close look at what the vendors are doing, the SFC can
-help them.
+Another activity is related to defending the license, which is the
+GPLv2. For example, there have previously been vendors distributing
+Git with some changes, but without providing the source code for the Git version they
+were distributing. So far it has been possible to resolve these cases, but
+it is not completely clear if all vendors are currently fullfilling all of
+their obligations. If any developers who have contributed to Git
+want to take a closer look at what the vendors are doing, the SFC are
```
chriscool: I am not sure about "the SFC are". Isn't "is" better than "are". I guess it depends if we are talking about the people in the SFC or the SFC itself.

chriscool:
Except maybe one small nit I commented about, it looks good to me, so I plan to merge it soon. Thanks @Cogito!

rev_news/drafts/edition-15.md
peff: I think "is" is better in this context, as it is talking about the group a a whole (see http://public.wsu.edu/~brians/errors/group.html, for example). I don't think either is egregiously wrong, though.

BTW, not a new problem in this PR, but a while ago Conservancy sent out some "style points" for referring to it (them? :wink:):

> - please do not add a "the" in front of "Software Freedom Conservancy" or "Conservancy" in any written context.  We've admittedly used a "the" intermittently in the past, but we'd like to be consistent going forward.  So, instead of "Project X is a member of the Conservancy," just say "Project X is a member of Conservancy."
>
> - please use "Conservancy" as the abbreviation for "Software Freedom Conservancy."  We're discontinuing the use of the acronym "SFC," and we ask your help in doing the same.

I don't think it's worth trying to change existing references, but perhaps something to keep in mind for future writing.

Cogito: I think 'is' is probably correct too, however as @peff points out, if we are not using the preferred style for referencing Conservancy I think we should fix it.

The usage seems odd to my ear, but I'm happy to go through and make the appropriate changes.

chriscool: @peff thanks for your input about "is" vs "are" and about Conservancy ;-)
@Cogito yeah it would be nice if you could fix all of that, thanks!

chriscool:
Great, thanks Cogito!

PR #159, opened 2016-06-09
convert all internal links to https

peff:
Now that GitHub Pages can use https (and we'll be upgrading http requests to https), we can use https links everywhere to avoid the upgrade step.

This is a purely mechanical conversion.

I do wonder if some of these could be relative links (they are all referring to the git.github.io site itself), but here I've taken the most conservative route.

/cc @git/web


chriscool:
LGTM, thanks!

PR #160, opened 2016-06-12
Ensimag contributing to git

aironefr:
Like in 2015 (https://git.github.io/rev_news/2015/07/08/edition-5/),
students from engineering school Ensimag contributed to Git.  This is
a small article to thank the community and to show what we did during
these 4 weeks.

We really wanted this article to be personal, so we wrote using
first-person pronouns.

It's quite long, so it could be moved to another part of the Rev News
and it's not really a "Discussion", so it doesn't really fit to the
first section.

### DIFF
--- a/rev_news/drafts/edition-16.md
+++ b/rev_news/drafts/edition-16.md
@@ -19,6 +19,22 @@ This edition covers what happened during the month of May 2016.
 
 ### General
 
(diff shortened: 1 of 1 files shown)


chriscool:
It looks great! Thanks guys!

PR #165, opened 2016-07-19
Stylistic change, to avoid duplicated "move" in a sentence

jnareb:
I think it makes the sentence easier to understand, especially to not native English speakers (there were "move" twice in a a single sentence, each with different meaning: move ~= change or action, and move ~= shift or push). This is not the only way of improving it.


chriscool:
Thanks Jakub!

PR #166, opened 2016-07-20
Expand "Making open source data more available" entry

jnareb:
Explain what "Making open source data more available" GitHub blog post is about. This should help readers decide if it is what they want to read.


tfnico:
Thanks! The links didn't get a lot of love this time due to time pressure on my side.

Issue #157, opened 2016-05-16
Broken Links in Git Rev News


gitster:
Introductory text of git.github.io/rev_news reads like so: """Git Rev News: Edition 15 (May 11th, 2016)
Welcome to the 15th edition of Git Rev News, a digest of all things Git. For our goals, the archives, the way we work, and how to contribute or to subscribe, see the Git Rev News page on git.github.io.

This edition covers what happened during the month of April 2016, especially at the Git Contributor Summit on April 4 2016 and at the Git Merge conference on April 5 2016."""

Among three links, two of them go non-existent pages ("Git Rev News", and "Git Rev News page").

chriscool:
@gitster thanks, these links are old style links that indeed do not work any more.
They are fixed in commit d1006a57529513e7bbbe4f8234b4dcb01b9460ce.

PR #167, opened 2016-07-22
Fix broken `Git Rev News page` links in Git Rev news

rtyley:
A few of those old `rev_news.html` links made it through... (after  https://github.com/git/git.github.io/issues/157). Also using HTTPS for preference.


tfnico:
Sounds good! Thanks!

PR #173, opened 2016-08-15
edition-18.md: Add the git-series announcement

joshtriplett:
Christian Couder requested that I submit a short item on git-series for the "Git tools and sites" section.


tfnico:
Thanks, Josh!

PR #177, opened 2016-08-17
Add light reading (VC for gamedev)

jnareb:
Link "Version control: Effective use, issues and thoughts, from a gamedev perspective"

_NOTE: this is from 02 August 2016, if it matters_


tfnico:
Thanks!

PR #193, opened 2016-10-18
rn20: expand a but about Dscho's work on interactive rebase

jnareb:
Add an information from Dscho about the use of Scientist library to ensure that the patches making interactive rebase faster are in fact correct.

I'm not sure if the new information is well connected to the rest of it; as it is now, this point has actually two endings...

### DIFF
--- a/rev_news/drafts/edition-20.md
+++ b/rev_news/drafts/edition-20.md
@@ -124,6 +124,29 @@ around the sequencer are of course very nice. It is interesting and
 satisfying to see that they are the result of building on top of
 previous work over the years by GSoC students, mentors and reviewers.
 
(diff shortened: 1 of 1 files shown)


chriscool:
@jnareb I think this could go into a short separate article just below the existing one on rebase -i.
Are you ok to change it into an article?

jnareb:
@chriscool I don't mind changing it into a (short) article. 

This would solve the problem of two endings ;-)

chriscool:
Ok I will do that then.

chriscool:
See https://github.com/git/git.github.io/commit/23f9abea01a9fd97b6f0904795cf20c9a6598592

PR #194, opened 2016-10-18
rn20: Git User's Survey 2016 ends soon

jnareb:
Just a reminder for those who have not answered yet.

The "Git User's Survey 2016" is to end at 20 October 2016... one day after Git Rev News edition 20 is planned to come out.


tfnico:
Thanks!

PR #195, opened 2016-10-18
rn20: Changing the default for "core.abbrev"?

jnareb:
Add discussion about changing the default abbrev length and dealing with abbreviated SHA-1s.

I'm not sure if I put it in the correct place; though as it is in 'next' now, perhaps "Reviews" and not "General" subsection of "Discussions" is a good choice. Note also that URLs to commits might need cleanups.

Should I have added a byline?

### DIFF
--- a/rev_news/drafts/edition-20.md
+++ b/rev_news/drafts/edition-20.md
@@ -24,6 +24,46 @@ This edition covers what happened during the month of September 2016.
 
 ### Reviews
 
+* [Changing the default for "core.abbrev"?](http://public-inbox.org/git/CA+55aFy0_pwtFOYS1Tmnxipw9ZkRNCQHmoYyegO00pjMiZQfbg@mail.gmail.com/)
+
+Linus Torvalds asked for increasing the default value for number of
+characters in SHA-1 abbreviations.  The `default_abbrev = 7` was
+reasonable at early days of Git, but the project of size of the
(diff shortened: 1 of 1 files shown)


chriscool:
Great, thanks @jnareb !

rev_news/drafts/edition-20.md
```diff
@@ -24,6 +24,46 @@ This edition covers what happened during the month of September 2016.
 
 ### Reviews
 
+* [Changing the default for "core.abbrev"?](http://public-inbox.org/git/CA+55aFy0_pwtFOYS1Tmnxipw9ZkRNCQHmoYyegO00pjMiZQfbg@mail.gmail.com/)
+
+Linus Torvalds asked for increasing the default value for number of
+characters in SHA-1 abbreviations.  The `default_abbrev = 7` was
+reasonable at early days of Git, but the project of size of the
+Linux kernel needs `git config --global core.abbrev 12`; while Git
```
chriscool: Also I wonder if the semicolon could be replaced by a full stop.

peff: Yes using "a" here is correct (and probably `s/at early days/in the early days/`).

I also agree that a period is preferable to a semi-colon here.

chriscool: Thanks @peff ! I will make those changes.

PR #196, opened 2016-10-19
rn20: musl libc vs NO_REGEX / REG_STARTEND

jnareb:
Add description of discussion about Git no longer working with musl libc's regex implementation out of the box (without setting `NO_REGEX=1` in build configuration).

I'm not sure if it is worth including, but I think it might be interesting (the relying on POSIX issue, and no code is easier maintenance). Also, perhaps it could be in "Support" instead.

Might need style tuneups, and perhaps byline.

### DIFF
--- a/rev_news/drafts/edition-20.md
+++ b/rev_news/drafts/edition-20.md
@@ -17,9 +17,56 @@ This edition covers what happened during the month of September 2016.
 
 ## Discussions
 
-<!---
+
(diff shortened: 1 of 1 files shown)


chriscool:
@jnareb great, thanks again!

chriscool:
Actually I will move it into the "Support" section as it seems better given the title.

PR #197, opened 2016-10-19
rn20: Discussion around git-log-compact

jnareb:
Add description of discussion leading to git-log-compact project (originally proposed as `contrib/git-log-times` script).

This time with byline...


chriscool:
Great, thanks again!

PR #200, opened 2016-10-31
Fix 404 on About in Navbar

vsmart:
Currently "about" in navigation bar is a 404. This is a fix - now it renders `about.md` as expected.


peff:
Thanks!

PR #201, opened 2016-11-05
Rev News: mention th Git poem

dscho:


tfnico:
Thanks! That was good :)

PR #211, opened 2016-12-16
Add link to gitql tool (git query language)

jnareb:


tfnico:
@jnareb I wouldn't mind if you just commit and push these directly, if you prefer.

PR #219, opened 2017-01-26
microprojects 2017

pranitbauva1997:
Removes ideas which have been done before and add one new microproject.


peff:
Thanks for kicking this off. Glancing over the projects, I think some of them are not very "micro" (like the XDG credential-cache one), but let's use this as a starting point.

stefanbeller:
Thanks for starting the micro projects page!

PR #223, opened 2017-02-09
soc-2017 typographical fixups

peff:
This is a mixed bag of typographical fixups, along with converting references to gmane to point to public-inbox instead. I only bothered with the 2017 pages, as the older ones are basically historical. We should perhaps consider removing them from the navigation bar on the side (we could also remove them completely, but perhaps people would want to refer to them? They'd always be available in the history, of course).


peff:
I'm going to merge this immediately, as we don't know when GSoC folks will start looking at application materials. Comments or additional fixups welcome.

PR #224, opened 2017-02-09
navbar fixups

peff:
This tries to de-bloat the navbar a bit by moving historical summer of code materials to their own page. It also fixes a bug in our navbar generation which caused a blank entry.

### DIFF
--- a/Outreachy-2016-May.md
+++ b/Outreachy-2016-May.md
@@ -1,6 +1,7 @@
 ---
 layout: default
 title: Outreachy May 2016
+navbar: false
 ---
 
 # Outreachy May 2016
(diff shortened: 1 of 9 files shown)


chriscool:
This looks good, but it looks like there is still a blank entry below "Home" and "About". Not sure if it is by design or not.

git-developer-page

peff:
That was _supposed_ to be fixed https://github.com/git/git.github.io/pull/224/commits/c7afa1111e64c0b65c4e481c00ea21ac6a405eff, but yeah, I noticed it does not seem to have worked.

I suspect the problem is that the title is `nil` and not empty. Unfortunately since it doesn't happen on my local build, I have no way to test except by merging attempts to master. I'll try the `nil` thing, though.

peff:
Yep, that fixed it (545b8fc336347ff451bebec71b72afe70cf99f69).

chriscool:
Great, thanks!

PR #225, opened 2017-02-10
SoC-2017-Ideas: remove libgit2

chriscool:
As it looks like libgit2 people have not responded to emails we sent, we can assume that they don't want to participate to the SoC this year.

The mailing list thread is:

https://public-inbox.org/git/vpq1svtstud.fsf@anie.imag.fr/


chriscool:
@moy and @peff are you ok with this?

moy:
Yes I am.

chriscool:
Ok, thanks @moy!

PR #229, opened 2017-02-21
rn-24: fix broken markdown bullet list

moy:
Turn spaces into tabs to avoid rendering two items as code.


mjaix:
Does not seem to hurt on my box ... anyhow, the other bullets on this level have their lines started with tabs.

moy:
I noticed it because the rendering was broken on GitHub. Not sure it would still have been after being processed by Jekyll, but it was easy to fix. Thanks for merging.

PR #233, opened 2017-02-24
Rev News #24: mention Ben Peart, not his nephew

dscho:
... assuming here that Ben has a nephew called "Bob"... ;-)


tfnico:
Thanks!

PR #235, opened 2017-03-09
"Previous branch" is too big for microproject

gitster:
As I said in the review of Kannan's attempt in https://public-inbox.org/git/xmqqtw882n08.fsf@gitster.mtv.corp.google.com/, this is not a suitable topic for a micro-project, as doing it in a "micro" way adds more problems than it solves, and doing it correctly (which Kannan's attempt is a step in the right direction) is way more than a "micro".

Let's remove this.

### DIFF
--- a/SoC-2017-Microprojects.md
+++ b/SoC-2017-Microprojects.md
@@ -202,12 +202,6 @@ See the commit
 [c6f44e1da5](https://github.com/git/git/commit/c6f44e1da5e88e34)
 for example.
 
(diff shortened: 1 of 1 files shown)


peff:
Yes, I'd agree.

I was going to wait for an ack from the original author, but this comes (indirectly) from you (in 13796b45c61a223bd7ef647729df2a028a7701e7, and then copied from year to year). So I'll take that as an ack. :)

PR #250, opened 2017-06-16
rn-28: Add more places with inline code style - formatting fixes and improvements

jnareb:
Use `inline code` style to avoid turning '--' (double dash) in
description of long options into '–' (long dash) ligature in few
more places.  Balance backticks.

Add block of code formatting in more places, and provide explicit
programming language name for each code block.

Align start of line comments in code blocks.


chriscool:
This LGTM but I haven't tested how it looks in practice.
@jnareb feel free to merge this and then please check that it looks good on the web site.
Thanks!

PR #265, opened 2017-11-21
rn-33: change CRLF article end

chriscool:
Maybe this end better reflects the conclusion of the discussion.


jrn:
Yes, this looks like an improvement. Thanks for writing it.

jrn:
I think this is trying to summarize https://public-inbox.org/git/alpine.DEB.2.21.1.1710271716330.6482@virtualbox/.

As you say, in the presence of "* -text", the "*.sh text eol=lf" line is only about adding additional protection against messing up the line endings of .sh files, so it is not the end of the world that the pattern misses some other shell scripts.

chriscool:
Yeah, I am trying to summarize both of these emails.
As it seems better I will merge this.

PR #283, opened 2018-03-20
Rev News: mention that v2.17.0-rc0 is available also for Windows

dscho:


chriscool:
Thanks again!

PR #284, opened 2018-03-21
RevNews: add a link to a good explanation of Git's history simplifica…

dscho:
…tion

From that website:

	Git history simplification can be a confusing beast. 99% of the
	time you won't even know it exists, but every once in a while it
	will jump out of the dark corners of Git and bite you. In this
	article we'll explore what history simplification is and how it
	can cause confusion when looking at file history.


mjaix:
@dscho: Thanks!

PR #292, opened 2018-05-14
fix name typo

prertik:
and Karki is pronounced as "Car Key" 😃


chriscool:
Thanks @prertik and sorry about that!

PR #293, opened 2018-05-14
Add a few details about optimizing writes to unchanged files

newren:
In regards to the story about optimizing writes to unchanged files:

1) The text noted that en/rename-directory-detection-reboot would be
   merged to next soon (that's what the last cooking email said), but it
   has since been merged to next..
2) It's worth noting that Junio found a bug with Linus' proposed "stupid
   brute force" patch (namely that if a file involved in a rename has
   dirty changes before a merge, those dirty changes could be overwritten
   without warning to the user).
3) Finally, since there is a specific commit message that explains the
   history of the can-we-skip-worktree-updates check and what fix was
   used, link to it so the curious can dig more (without having to read
   all the patches about directory rename detection).

### DIFF
--- a/rev_news/drafts/edition-39.md
+++ b/rev_news/drafts/edition-39.md
@@ -93,10 +93,11 @@ approaches.
 Elijah also replied to Linus' alternative patch by discussing
 different approaches. And Junio agreed with the direction Elijah was
 taking, though he had not as much time as he would have liked to think
-this through at that time.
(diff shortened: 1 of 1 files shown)


newren:
I tried to match style and rewrite as little as possible while adding a few extra pointers.  Feel free to take whichever pieces make sense to you and modify as you see fit.

chriscool:
Your changes looks great! Thanks!

PR #296, opened 2018-06-19
clarify purpose and context of git branch workflow tools

aspiers:
Expand on the purposes of the tools, and mention that these are being announced due to the recent presentation to the Git London User Group.


chriscool:
Great, thanks!

PR #300, opened 2018-06-21
RevNews: fix misspelled name

dscho:
Ed's surname really has no 'p' in it: http://www.radiotfs.com/Show/141/ThereisNoPinThomson


chriscool:
Thanks @dscho and sorry @ethomson!

PR #301, opened 2018-06-21
RevNews: do not claim that the Microsoft/GitHub deal is final

dscho:
It has been made very clear from both Microsoft and GitHub that they announced an *agreement*. So the intent is there, but the acquisition is far from final, and therefore the companies are still independent of each other, and are still managed independently as long as the deal is closed.

We should not spread inaccurate news, so let's just state the facts correctly (otherwise this RevNews item could be misconstrued to be crafted to help competing companies by using that age-old despicable technique of spreading Fear, Uncertainty, and Doubt). There is no need to invent anything here, what I did was a slightly edited copy/paste from Satya's blog post.


chriscool:
Thanks @dscho again!

PR #302, opened 2018-07-07
Indent fix

Hendis:
fix indent in GSoC article


chriscool:
Shouldn't both the git:indent-fix and the Hendis:indent-fix branches should be rebased first?

chriscool:
Sorry if it was good before and I forgot to merge it.

PR #306, opened 2018-07-18
Fix SHA-1 size

lucaswerkmeister:
A SHA-1 hash has 160 bits (40 hex digits), not 128.

---

I really hope I’m not being stupid here and missed something obvious :)


chriscool:
Yeah, sorry about the mistake and thanks both!

PR #315, opened 2018-10-25
rev news sources: wrap long lines; fix broken link

waldyrious:
The old link http://thechangelog.com/tagged/git/ now redirects to https://changelog.com/tagged/git, which is a 404 page. This PR updates it to the new URL format.

In addition, some lines were wrapped to make the code view (and diffs) more readable. Rendered text is not affected.


chriscool:
Great, thanks!

PR #317, opened 2018-11-19
Very small typo in Edition 45

derrickstolee:
path -> patch


chriscool:
Thanks @derrickstolee !

PR #322, opened 2018-11-22
Fix gerrit release note links

dpursehouse:
The Gerrit Code Review project recently migrated its homepage from being hosted on googlesource.com and rendered by Gitiles, to being hosted on Firebase and rendered by Jekyll.

As a result of this migration, the homepage is still under the same domain, but the URLs for the release notes have changed. Instead of  being located at`/releases/$version.md` they are now located at `/$version.html`.

This PR changes the URLs accordingly, and also adds missing anchors for some of the maintenance releases, so that clicking on the link now takes the reader directly to the relevant section for that release.


chriscool:
Greqat, thanks!

PR #323, opened 2018-11-27
Add link and words for sr.ht

msingle:
Checked with the author (@ddevault) and they're ok with spreading the word about sr.ht.  
Text partially copied from announcement, but that's because that description's better than anything I could come up with.


ddevault:
Thanks!

PR #324, opened 2018-11-30
Rev News: mention Adobe's `git-server`

dscho:
Pretty interesting node.js-based application serving Git repositories via `http` and `https`.


chriscool:
Thanks @dscho !

PR #350, opened 2019-01-22
rn-47: Mention JGit and Gerrit security vulnerabilities

dpursehouse:


chriscool:
Do you have links so that interested people can take adeeper look?

dpursehouse:
There were no bug reports on jgit, only on gerrit. I can add links to
those, but they don't say much more than what I added here and what is
mentioned in the gerrit/jgit release notes that are already linked.

All the discussions around the bugs were done on the gerrit maintainers
mailing list which is not public.

chriscool:
I think it's still better to add the links even if they don't say much more.

dpursehouse:
@chriscool I've added another commit with the links.

chriscool:
Great! Thanks!

PR #355, opened 2019-02-11
Fix my name, add description for one of projects

telezhnaya:


chriscool:
Thanks Olga!

I think I will improve on your second patch a bit, but it is a great start!

PR #357, opened 2019-02-12
SoC-2019-Microprojects: remove "Move ~/.git-credential-cache to > ~/.cache/git"

MatheusBernardino:
This microproject was already solved at 60759ba ("credential-cache: use XDG_CACHE_HOME for socket", 2017-03-17).

This PR also removes all trailing spaces.


peff:
Thanks! I think the whitespace cleanup should be fine; other people may be touching the project list in preparation for GSoC, but the changed lines are all in the general description.

PR #359, opened 2019-02-25
remove stash'ing merge conflicts project

tgummerer:
This project depends on stash-in-c going to at least `next`.  I'm
going to send an updated version of the patch series tonight, but
I'm not very confident at this point that it's actually going to
make it to next before GSoC starts.  I'd rather take the project
away now, before applications actually open, and later
re-introduce it, than misleading a student.

---

I wish I didn't have to make this last minute change, however I think it's best not to propose a project that we're not sure at this point is available during summer.  I'd be more than happy to put it back onto the list once stash-in-c made it to `next`.

cc: @chriscool @telezhnaya


chriscool:
Ok, I agree that it's better to remove it for now. Thanks!

PR #366, opened 2019-03-10
SoC-2019-Ideas: don't split text within the `s

sivaraam:
When the text within the ` (tildes)  is split between two lines
it is not rendered as expected. To be precise the part that
comes in the second line isn't visible.

It's worth noting that this is not clear from the rendering
of the markdown by GitHub. But it's clear from the syntax
highlighting done by the editor in GitHub and the rendering
in the website (where this was first noted).

### DIFF
--- a/SoC-2019-Ideas.md
+++ b/SoC-2019-Ideas.md
@@ -214,9 +214,9 @@ This would consist in taking care of the following issues:
 * Bonus: Make a flag to allow rebase to rewrite commit messages that
   refer to older commits that were also rebased.  (i.e. if rebasing
   commits A and B, and commit B says `This reverts commit <sha-of-A>`,
-  then rewritten B's commit message should say `This reverts commit
-  <sha-of-rewritten-A>`.)  Do this for both sha1sums and sha1sum
-  abbreviations in commit messages.
+  then rewritten B's commit message should say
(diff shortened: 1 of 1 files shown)


chriscool:
Thanks again!

sivaraam:
You're welcome! :smile:

Actually, I also noticed another issue which I was thinking of fixing but couldn't find how to fix it. The issue is that the code block in the "git log --oneline improvements" section renders badly in smaller screen. Further it also breaks the single-pagedness of the website. As a consequence, one might have to zoom in to read the post and scrolling might take you to empty space in the right. Screen shots:

* Desktop view:
    Screenshot from 2019-03-10 20-56-49

* Mobile view (iPhone SE resolution)
    Screenshot from 2019-03-10 20-56-14

    Screenshot from 2019-03-10 21-03-46

    Note the scroll bar in the bottom of the page.

Are you aware of any possible way to fix this? I was thinking that a fix should make the code block alone to be scrollable rather than the whole page.

chriscool:
No, I am not aware about how we could fix it. It would be nice though if you could copy everything about it into a new issue, so that we could be reminded that this issue exists and maybe fix it one day.

sivaraam:
Done. #367

PR #372, opened 2019-03-22
SoC-2019: add two more micro project ideas

dscho:
These are kind of "papercuts" I am collecting, and rather than doing
so in my private repository, I might just as well put them into a
more useful spot.

### DIFF
--- a/SoC-2019-Microprojects.md
+++ b/SoC-2019-Microprojects.md
@@ -218,6 +218,28 @@ that we have `struct dir_iterator` (see dir-iterator.h), convert these
 to use the dir-iterator to simplify the code. Do only one conversion
 per microproject.
 
+### Let `git fetch <remote> HEAD` use the current branch name
+
+When fetching from a remote, one can specify which ref(s) to fetch.
+There is no convenient way to specify the current branch, though, and
+`HEAD` may be a good shortcut. Or `@`. Or `-`.
+
+### Teach `git range-diff` new tricks
+
+When contributing patches to Git, developers often need to compare
+in what form those patches made it upstream (e.g. to pick up fixes
+made by the Git maintainer in preparation for a new iteration). In
+such a case, a `--left-only` (or `--right-only`) option would come
+in real handy, to allow `git range-diff --left-only <branch>..upstream/pu`
+to skip commits that are upstream only (as there will be a metric
+ton of them).
+
+Another neat trick that could come in real handy at times would be
+to teach `git range-diff` to compare a local branch to a patch
+series that is provided as an mbox, e.g. to use `git range-diff` to see
+whether a given patch series on the mailing list has made it into
+git.git's `pu` branch yet.
+
 ## How to find other ideas for microprojects
 
 If you don't like for some reason the above microprojects or if you


SoC-2019-Microprojects.md
```diff
@@ -218,6 +218,28 @@ that we have `struct dir_iterator` (see dir-iterator.h), convert these
 to use the dir-iterator to simplify the code. Do only one conversion
 per microproject.
 
+### Let `git fetch <remote> HEAD` use the current branch name
+
+When fetching from a remote, one can specify which ref(s) to fetch.
+There is no convenient way to specify the current branch, though, and
+`HEAD` may be a good shortcut. Or `@`. Or `-`.
```
chriscool: I think `git fetch <remote> HEAD` already works and it might be difficult to convince people that it is ok to make it do something else than what it is currently doing. Or I am missing something?

SoC-2019-Microprojects.md
```diff
@@ -218,6 +218,28 @@ that we have `struct dir_iterator` (see dir-iterator.h), convert these
...
+
+### Teach `git range-diff` new tricks
+
+When contributing patches to Git, developers often need to compare
+in what form those patches made it upstream (e.g. to pick up fixes
+made by the Git maintainer in preparation for a new iteration). In
+such a case, a `--left-only` (or `--right-only`) option would come
+in real handy, to allow `git range-diff --left-only <branch>..upstream/pu`
+to skip commits that are upstream only (as there will be a metric
+ton of them).
```
chriscool: I guess `git range-diff --right-only <branch>..upstream/pu` would skip commits that are only in <branch>. I wonder if the syntax could be `git range-diff --left-only <branch> upstream/pu` (so without `..`) though. I also wonder if  `git range-diff --left-only <branch>...upstream/pu` (so with 3 dots) or `git range-diff --left-only <base> <rev1> <rev2>` would make sense?

SoC-2019-Microprojects.md
```diff
@@ -218,6 +218,28 @@ that we have `struct dir_iterator` (see dir-iterator.h), convert these
...
+
+Another neat trick that could come in real handy at times would be
+to teach `git range-diff` to compare a local branch to a patch
+series that is provided as an mbox, e.g. to use `git range-diff` to see
+whether a given patch series on the mailing list has made it into
+git.git's `pu` branch yet.
```
chriscool: This idea seems a little big for a microproject to me as I don't think range-diff can process patches right now.

SoC-2019-Microprojects.md
dscho: If by "already works" you mean to say "does something that is utterly incomprehensible and leaves the user totally puzzled", then I absolutely agree.

If you don't mean that, then I respectfully rescind my agreement with you.

dscho: > If by "already works" you mean to say "does something that is utterly incomprehensible and leaves the user totally puzzled", then I absolutely agree.

(I assume you know what `git fetch origin HEAD` is doing???)

SoC-2019-Microprojects.md
dscho: The way `git range-diff` parses its option, all three invocations (1, 2 and 3 arguments all are translated into the two arguments `A..B C..D`) boil down to the same. So yes, it would work. It would actually be hard to make it *not* work for all three invocations.

SoC-2019-Microprojects.md
dscho: Hey, this was a challenge for you, maybe you missed it... Just have a look at the implementation of `git range-diff`.

Here, I'll help you: https://github.com/git/git/blob/v2.21.0/range-diff.c#L27-L148

As you should have seen: `range-diff` simply takes the commit ranges, throws them at `git log -p` and then parses the output.

And *obviously* there is not a whole lot of work to do if you want to teach `range-diff` to read an mbox instead. We already have pretty much all the code that is needed, it just needs to be stitched together appropriately.

SoC-2019-Microprojects.md
chriscool: Ok, nice!

SoC-2019-Microprojects.md
chriscool: Anyway I think a micro-project should be something more straightforward than that. At least I think the description should include something like "this means replacing the current behavior". I can add that if you want.

SoC-2019-Microprojects.md
chriscool: I still feel that it is on the big side for a micro-project but I am ok to merge this anyway.

SoC-2019-Microprojects.md
moy: > I assume you know what `git fetch origin HEAD` is doing???

From a quick test, I'm guessing that it fetches the remote ref named `HEAD` into `FETCH_HEAD`, which is what I had expected. Did I miss something (very likely so actually ;-) )?

dscho: > From a quick test, I'm guessing that it fetches the remote ref named `HEAD` into `FETCH_HEAD`, which is what I had expected.

Technically, you are correct. Practically, the way `HEAD` is updated in a remote repository is... unintuitive. For example, last time I checked, when I changed my default branch on GitHub, it did not change `HEAD`. Other hosters like repo.or.cz do not even *have* the concept of a default branch. Git itself does not have any concept of a remote repository's default branch at all, it always defaults to cloning `master`.

Hence my statement that it is totally unintuitive what this does (and confusing).

dscho: > At least I think the description should include something like "this means replacing the current behavior". I can add that if you want.

Sure, but let's add something that describes this a bit better. If you say that we change existing behavior, it kind of implies that the current behavior is well-defined: but it is not.

moy: OK, I mostly agree with you, but I think the confusing part is the very notion of `HEAD` in a remote repository, not really what `git fetch HEAD` does in this situation. I'd find it confusing if `git fetch remote master` was referring to the *remote* branch master, and `git fetch remote HEAD` to the *local* ref `HEAD`: to me the ref argument of `git fetch` is always a remote ref. 

> Git itself does not have any concept of a remote repository's default branch at all, it always defaults to cloning master.

That one is not true, though: `git clone` does clone the branch pointed-to by the remote's `HEAD` (modulo a "bug" when multiple branches point to the same commit, fixed in 1.8.4.3).

`-` would be very confusing to me too, since it usually means "the branch that used to be checked out before this one", not "the currently checked-out one".

Using `@` would be the less confusing to me. Perhaps `git fetch --current` would be best actually: no possible confusion, and explicit about what it would do.

moy: > For example, last time I checked, when I changed my default branch on GitHub, it did not change `HEAD`.

Just tested out of curiosity, and it did work for me: `git ls-remote --symref` shows the branch set as "default branch" on GitHub.

SoC-2019-Microprojects.md
dscho: If students are left entirely to their own devices, sure, it might be on the big side. But the point of GSoC is not cheap labor, is it? It is a student/mentor setup, and there is no good reason why that relationship should only start *after* the micro-project. Unless the mentor wants to avoid work, that is, in which case one might ask why they volunteered to mentor in the first place.

SoC-2019-Microprojects.md
dscho: @moy ?

moy: I [already explained](https://github.com/git/git.github.io/pull/372#discussion_r268823051) why I thought `-` would be confusing and inconsistent with other commands. `git fetch @` would probably be OK, though I'm not sure about all the uses of the `@` shortcut in Git.

Personally, I wouldn't use this feature, I always fetch everything so `git fetch` without argument is perfect for me. I guess people working on huge repositories may want to fetch only the current branch if only to gain a bit of download time, but I suspect this would be a relatively rarely used feature, hence wouldn't need a short syntax. If so `git fetch --current` would be best IMHO: no ambiguity, and not that long to type (especially if you have smart completion, `--c<TAB>` is just 4 keys).

dscho: > `git fetch --current` would be best IMHO: no ambiguity, and not that long to type (especially if you have smart completion, `--c<TAB>` is just 4 keys).

:+1:

dscho: Or `--current-branch`...

dscho:
Okay, since GSoC 2019 is already in almost full swing, I'll close this PR unmerged. I moved the corresponding ideas to

* https://github.com/gitgitgadget/git/issues/205
* https://github.com/gitgitgadget/git/issues/206
* https://github.com/gitgitgadget/git/issues/207

PR #373, opened 2019-04-01
add Thomas as possible mentor for a couple of projects

tgummerer:
@dscho convinced me privately that mentoring these projects by myself would be doable.  Add myself as potential mentor for them.


chriscool:
Great, thanks!

PR #377, opened 2019-05-09
Update issue templates

MrJPELAEZ:


Andrey0777:
Здравствуйте  or Hello!

chriscool:
It might be a good idea for big software projects developed on GitHub to have an issue template, and the one you suggest looks nice for such a project. This project though is just about developing a relatively small website, and I don't think it deserve such an issue template.

I am perhaps wrong, but I think it could send the wrong message to Git contributors who are used to send an email to the mailing list when there is a bug in Git.

I also don't think we have had bad bug reports about the website.

So I would be in favor of closing this without merging.

Thanks anyway for the merge request and the suggestion! Maybe one day we will need something like this and it could be useful then.

chriscool:
As there has been no further discussion, I think this can be closed. Of course I am ok with reopening this if we think we could benefit from an issue template.

jpelaezClub:
Update README.md

PR #383, opened 2019-06-26
Rev News: fix the Git for Windows version

dscho:
Git for Windows' versioning scheme _drops_ the `(1)` for the first version that is based on an upstream version. I have fixed this many times, and will hopefully not have to fix it many more times.


chriscool:
Thanks, merged!

chriscool:
@dscho please note that the "Releases" section is generated mostly automatically using this script: https://github.com/chriscool/getreleases/blob/master/get_releases.py
I will see if I can fix the script soon.

PR #388, opened 2019-08-10
Proposition of subject for translated manpages in rev 54

jnavila:
I don't really know how long this can be. What kind of details may be interesting?

Any feedback welcome.


chriscool:
I think it's ok like that, or if you want you can make it longer and put it in the Discussions -> General section. Thanks a lot!

chriscool:
Merged, thanks!

PR #389, opened 2019-08-19
edition 54: Developer Spotlight

jnavila:
A quick proofreading would be great.


chriscool:
Great! We will proof read it later. Thanks!

PR #392, opened 2019-08-21
rn-54: update links to corrected docs, fix sed command

newren:
Found a few small typos and errors.  The most embarrassing one was the sed command in the one example I gave was broken (serves me right for changing my real module name to 'my-module' and then not re-testing), but some of the documents I linked to had a few minor issues too, so I updated them and made this PR to link to the newer versions.


chriscool:
Merged, thanks!

PR #397, opened 2019-09-23
Fix spelling of James Ramsay

jamesramsay:
Very small typo 😄


chriscool:
Sorry about the typo James and thanks for the fix!

PR #402, opened 2019-10-24
edition-56.md: proofread four  words to help out

Georgecanfly:
I noticed measured and acknowledged misspelled twice. Thought it would be nice practice to proofread and create a pull request. 

Signed-off-by: george espinoza <gespinoz2019@gmail.com>


chriscool:
@Georgecanfly yeah we (actually @mjaix does it) usually proofread before publishing the edition.

Thanks!

PR #406, opened 2019-11-18
Few suggested fixes to the 57th edition

sivaraam:


rev_news/drafts/edition-57.md
```diff
@@ -150,7 +150,7 @@ This edition covers what happened during the month of October 2019.
   `git diff` that would allow apps like `tig` and `magit` to show
   moved lines.
 
-  I'd also like to improve git push and pull to extend the idea of
+  I'd also like to improve Git push and pull to extend the idea of
```
chriscool: I think here it's ok to use "git". Perhaps we should even have back-ticks around "git push" and around "git pull" too.

chriscool:
Except for the comment I left, everything looks good! Thanks!

rev_news/drafts/edition-57.md
sivaraam: I was on the fence with that particular instance. I thought it best to change it and let some else point out if its wrong 😉

chriscool:
Thanks! Merged!

PR #413, opened 2020-02-14
RevNews: add an important heads-up for SSH users

dscho:
Technically, it is not a piece of _Git_ news per se, but IMO it is rather important for many Git users to be made aware of the deprecation (and soon, disabling) of the `ssh-rsa` keys.


chriscool:
Thanks Dscho!

PR #414, opened 2020-02-16
General-Application: add note about giving back and mentoring

chriscool:
Let's encourage GSoC students and Outreachy interns to contribute and to mentor.

### DIFF
--- a/General-Application-Information.md
+++ b/General-Application-Information.md
@@ -40,7 +40,8 @@ the result of your thinking as a review.
 
 A complete application should include a presentation of yourself
 (include any argument that may convince mentors that you are able to
-complete the project) and detailed explanations about your project.
+complete the project) and detailed explanations about your
+project. See also the "Note about giving back and mentoring" below.
 
 Project ideas are just ... ideas! The list we provide is not
 exhaustive, and more importantly each idea only includes a summary of
@@ -121,6 +122,39 @@ discussions.)
 The Git organization usually has very limited mentoring capacity.
 These days we usually accept between 0 and 3 students per season
 (Winter or Summer).
+
+## Note about giving back and mentoring
+
+We appreciate very much students and interns who stay around after the
+mentoring period is over. It is very nice to see them on the mailing
+list, even if they don't contribute much. It's of course better when
+they continue to contribute though, even by just reviewing a patch
+from time to time.
+
+Some people have been around for more than 10 years, others have
+become regular contributors and that's great!
+
+One very nice way to contribute and to give back is to mentor or
+co-mentor other students or interns coming after you. It helps create
+more opportunities for more students and interns like you, as
+mentoring capacity is the main factor preventing us from accepting
+more students and interns. If each student or intern accepted to
+co-mentor twice (once in Summer and once in Winter) just after they
+have been mentored, our mentoring capacity could increase
+significantly each year.
+
+Unfortunately very few former students or interns have been willing to
+just co-mentor along with an experienced mentor, even though it
+doesn't require much work, technical ability or time.
+
+Other free or open source projects have done better than us on
+this. At the Google Summer of Code Mentor Summit for example, more
+than 30% of the mentors that are former students.
+
+Therefore if you can show us in your application that you like
+mentoring, giving back and sticking to the projects you have worked
+on, you are likely to significantly increase your chances to be
+selected.
 
 ## Note about refactoring projects versus projects that implement new features
 


General-Application-Information.md
```diff
@@ -122,6 +123,39 @@ The Git organization usually has very limited mentoring capacity.
...
+more opportunities for more students and interns like you, as
+mentoring capacity is the main factor preventing us from accepting
+more students and interns. If each student or intern accepted to
+co-mentor twice (once in Summer and once in Winter) just after they
+have been mentored, our mentoring capacity could increase
+significantly each year.
+
+Unfortunately very few former students or interns have been willing to
+just co-mentor along with an experienced mentor, even though it
+doesn't require much work, technical ability or time.
```
tgummerer: I don't feel like this is quite right.  Being a good mentor does require work and time, and I think a mentor should be able to invest a few hours a week for mentoring.  Depending on the student that may not always be necessary, but every mentor should be able to dedicate that time.  For some people that might not seem a lot of work, but for others it is.

I think it would be better to write down our expectations for mentorship, rather than a blanket "doesn't require much work".

Similarly, I think a mentor should have some technical ability, even though probably less is required from a co-mentor.  But they should be able to navigate the git codebase, and help a student find their way through.  I think a lot of students are able to do that after completing GSoC or Outreachy, but I'd rather phrase it as "Successful former students/interns usually have adequate technical ability to be a successful mentor".

General-Application-Information.md
```diff
@@ -122,6 +123,39 @@ The Git organization usually has very limited mentoring capacity.
...
+
+Other free or open source projects have done better than us on
+this. At the Google Summer of Code Mentor Summit for example, more
+than 30% of the mentors that are former students.
+
+Therefore if you can show us in your application that you like
+mentoring, giving back and sticking to the projects you have worked
+on, you are likely to significantly increase your chances to be
+selected.
```
tgummerer: I don't think we should judge an application based on that.  It's easy to write that in an application, and we've seen students be interested in continuing to be part of the community, but life gets in the way.  I'd like for more students to stick around, but I don't think we can require that, especially as GSoC is specifically scoped to a summer, and the students don't have any obligations after that.

I'd rather try to find out the reason why students don't stick around afterwards, and try to eliminate obstacles to that, than making them show this in the application, which is easily faked.

General-Application-Information.md
chriscool: Yeah, you are right. I think it will be better with what you suggest.

General-Application-Information.md
chriscool: I think it's difficult to find the reasons why student or interns don't stick around. They can just say that they don't have much time because of their studies or work, and it's difficult to then go deeper.I also think it's important to have applicants reflect on that

sivaraam: I was also thinking of something along the lines of @tgummerer's thoughts. I particularly thought that the phrase "_...  you are likely to significantly increase your chances to be selected._" would make a student think that mentoring/co-mentoring after their intern as some kind of mandatory requirement to apply instead of something that would increase their chances of getting selected. This would be bad for students who don't have previous mentoring experience because I don't think they would be able to confidently mention whether they would be able to mentor after GSoC _while applying for internship_. These students might be able to answer that question with more confidence after their internship period. This _was_ my concern.

Your new suggestion:

> "Consider showing us in your application previous mentoring, giving back and community activities that you have done, especially related to free or open source software."

... sounds a lot better and I really don't have any concerns about this if this would replace the last paragraph 🙂

General-Application-Information.md
```diff
@@ -122,6 +123,39 @@ The Git organization usually has very limited mentoring capacity.
...
+significantly each year.
+
+Unfortunately very few former students or interns have been willing to
+just co-mentor along with an experienced mentor, even though it
+doesn't require much work, technical ability or time.
+
+Other free or open source projects have done better than us on
+this. At the Google Summer of Code Mentor Summit for example, more
+than 30% of the mentors that are former students.
+
```
sivaraam: Possible correction: ... more than 30% of the mentors were former students.

chriscool: Yeah, thanks for this suggestion!

General-Application-Information.md
tgummerer: Yeah I think that sentence is much better, thanks for rephrasing it!

chriscool:
I just created https://github.com/git/git.github.io/pull/418 to improve on this PR.

PR #415, opened 2020-02-18
rn-60: add interview with Chris Webster

sivaraam:
I've added the interview with Chris Webster with some minor fixes like correcting git -> Git etc.

One thing I'm not sure abuot is whether we have to explicitly mention that he is a contributor to GitGitGadget before the list of questions.


chriscool:
Thanks Chris and @sivaraam !

I think the first question makes it clear what Chris contributes to, so I don't think we need to add any mention before.

PR #416, opened 2020-02-19
rn-60: add note about suggestion of `git difftool` in interview

sivaraam:
Add a note that `git difftool` was suggested as a way to get side-by-side
diff in Git.


chriscool:
Thanks Kaartic and Chris!

PR #418, opened 2020-02-19
General-Application: rephrase note about giving back

chriscool:
Improve on https://github.com/git/git.github.io/pull/414 that was merged prematurely.

PR #420, opened 2020-02-26
General-Application: add quote about mentoring

chriscool:
This is an update following https://github.com/git/git.github.io/pull/418 and https://github.com/git/git.github.io/pull/414

cc @sivaraam @tgummerer

### DIFF
--- a/General-Application-Information.md
+++ b/General-Application-Information.md
@@ -152,6 +152,16 @@ Other free or open source projects have done better than us on
 this. At the Google Summer of Code Mentor Summit for example, more
 than 30% of the mentors were former students.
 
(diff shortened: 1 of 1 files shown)


sivaraam:
IMO, this quote does seem fine. Is there any reason for keeping this after the "Consider showing us in your application ..." paragraph? If not, it would be better to move this just above that paragraph as it seems like a more appropriate place for the quote to me.

chriscool:
@sivaraam and @tgummerer thanks for your reviews!

@sivaraam I agree the quote is better before the "Consider showing us in your application ..." paragraph, so I moved it there.

PR #429, opened 2020-04-21
a couple of typofixes

gitster:


chriscool:
Merged, thanks @gitster and @dscho!

PR #431, opened 2020-05-26
RevNews: mention isomorphic-git v1.0.0

dscho:
It's relatively old news by now (late February), but if other people are as behind on news because of COVID-19 as I am, they might appreciate the pointer, still.


chriscool:
Ok, thanks @dscho !

PR #437, opened 2020-06-07
Hacking-Git: highlight correct lines in CodingGuidelines

Denton-L:
The link to "Launching gdb explanations in CodingGuidelines" does not
actually highlight the bullet point about launching GDB. Correct the
hash fragment so that the correct lines are highlighted.

### DIFF
--- a/Hacking-Git.md
+++ b/Hacking-Git.md
@@ -26,7 +26,7 @@ people to interesting material that has already been written.
 
 * [Git for Windows' Debugging Git](https://github.com/git-for-windows/git/wiki/Debugging-Git)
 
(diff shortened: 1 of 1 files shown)


chriscool:
Thanks! I think though that the lines will have to be changed again, if some changes are made to the file in `master`. If we used a sha1 instead of `master`, the issue would be that the content we link to could become obsolete. I don't see a good solution to that.

chriscool:
Or actually if the section had a title with an anchor, we might use a link to that anchor.

Denton-L:
I don't anticipate that the GDB instructions will change very often so I don't think those instructions in particular will be stale any time soon. Since there's already a link to the `master` version of CodingGuidelines above, we should probably just hardcode the gdb link to point to a hardcoded hash. That way, we get the best of both worlds.

PR #446, opened 2020-07-26
rn-65: add interivew with Jonathan Tan

sivaraam:
I've added some links where I felt appropriate. One thing I'm not sure is if it would be nice to link "JGit" to https://www.eclipse.org/jgit/ or not.

### DIFF
--- a/rev_news/drafts/edition-65.md
+++ b/rev_news/drafts/edition-65.md
@@ -29,9 +29,58 @@ This edition covers what happened during the month of June 2020.
 ### Support
 -->
 
-<!---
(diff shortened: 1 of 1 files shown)


chriscool:
Thank @sivaraam! I think it's ok to add a link to JGit. I often add some links and then ask the person who was interviewed what they think about the changes I made.

sivaraam:
> 
> Thank @sivaraam! I think it's ok to add a link to JGit. I often add some links and then ask the person who was interviewed what they think about the changes I made.

That makes sense. I'll inform Jonathan about this.

PR #447, opened 2020-07-28
rn-65: add The history of `master` in git article

Cogito:
Here is my initial attempt at an article. Please let me know your thoughts and what can be improved.

The 'General' section seemed most appropriate as I referenced quite a few different posts to the mailing list thread to show how the discussion evolved.

I wasn't sure about quoting style that should be used, and I ended up using quote formatting for the main arguments to try and highlight them, in-lining other quotes where it seemed appropriate.

I wasn't entirely sure how to resolve the article, so any feedback on that would be great as well.

### DIFF
--- a/rev_news/drafts/edition-65.md
+++ b/rev_news/drafts/edition-65.md
@@ -17,11 +17,94 @@ This edition covers what happened during the month of June 2020.
 
 ## Discussions
 
-<!---
 ### General
--->
 
-
+* The history of `master` in git
+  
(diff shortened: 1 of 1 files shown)


chriscool:
Thanks! The article and its resolution look good to me! It's also the right thing to put it in the 'General' section as you did.

chriscool:
I made some small improvements (mostly about using "Git" instead of "git") in d0fbaf2df8447d67b08f75b0ee3410a796b991cf.

PR #448, opened 2020-07-28
Update edition-65.md

carenas:
typo?


chriscool:
@carenas thanks for PR! I actually wanted to say "series" and I realized that the end of the article has to be changed now that Git 2.28.0 has been released, so I hopefully fixed all this in 7346429683054f4163b7dfe1492f30393da3e114 instead of merging this PR.

PR #449, opened 2020-07-29
Created CODE_OF_CONDUCT.md

jivthesh:


peff:
Since this repo is really an extension of the Git project, would it make sense to use the CoC from git.git?

PR #454, opened 2020-07-30
Created LICENSE

jivthesh:


peff:
I don't mind adding a license file or otherwise clarifying the IP rules around content on this site. But does the GPL make sense for this mostly-text page? Perhaps one of the creative-commons licenses would be more appropriate.

jivthesh:
@peff so should ad a common license ?

peff:
I think a CC license would make more sense than GPL (or MIT would be fine, too). But in adding a license, I think it would make sense:

  1. To get input from people who add most of the content here, which is people working on revnews.

  2. We'd need agreement from people who added content already to grant the new license for that existing content. I imagine most people would be happy to do so (and I certainly would for my parts), but without that it creates a weird situation where some of the content has no license and some is under the new license.

sivaraam:
Closing this due to lack of progress. Feel free to open another one with the suggested changes :)

PR #457, opened 2020-09-19
micro: unify semantics for -dirty suffix around submodules

gitster:
A leftover bit


chriscool:
Thanks a lot @gitster !

PR #461, opened 2020-10-02
rn-67: fix wording and grammar

phil-blain:
I noticed when I read Rev News that the wording could be better for one sentence, and that the link I added to my script should have followed the example of the preceding links by including a verb. So here I am fixing that.


chriscool:
Thanks @phil-blain !

PR #465, opened 2020-11-15
Add light reading link: Git before GitHub

tarunbatra:

PR #466, opened 2020-11-16
rn-69: fix link to 'Git before Github' article

tarunbatra:
Fixes issue in the link caught by @sivaraam [here](https://github.com/git/git.github.io/pull/465#issuecomment-728202495).


chriscool:
Thanks @tarunbatra for the fix! And thanks @sivaraam for noticing!

Issue #463, opened 2020-10-30
Any comment about upcoming Git Rev News edition 69


chriscool:
A currently mostly empty draft is there:

https://github.com/git/git.github.io/blob/master/rev_news/drafts/edition-69.md

Feel free to comment in this issue, suggest topics, suggest persons to interview, or use the edit button (that looks like a pen) to edit and create a pull request with the changes you would like.

Let's try to publish this edition on Wednesday November 25th 2020!

Thanks!

cc @jnareb @mjaix @sivaraam @gitster

sivaraam:
@chriscool When do we plan to publish this edition? November 18th or November 25th? 🤔

chriscool:
@sivaraam I meant November 25th as written in https://github.com/git/git.github.io/blob/master/rev_news/drafts/edition-69.md but I had forgotten to update the comment when I copy pasted it from the previous issue. It's fixed now. Thanks!

jnareb:
I think it might be good idea to add a short note about planned deprecation and later removal of old `.git/branches/` and `.git/remotes/` way of specifying remote repositories, see <https://public-inbox.org/git/cover.1494509599.git.johannes.schindelin@gmx.de/t/#u>

chriscool:
@jnareb yeah, let's do that!

PhilipOakley:
"Git is too Hard"

An interesting blog/podcast about the Git CLI, for those who work inside the terminal https://changelog.com/posts/git-is-simply-too-hard

chriscool:
@PhilipOakley thanks! It was also on Hacker News.

jnareb:
I have added my links in 66b6960 - it includes links to "Git is too Hard" blog/podcast mentioned by @PhilipOakley .

chriscool:
Sorry I am very late. The script to automatically get the release information from many places broke for GitLab as they changed the format of their "releases" page and it is not so easy to fix.

mjaix:
Hi,

I tried to understand
why https://git.github.io/rev_news/2020/08/27/edition-66/, https://git.github.io/rev_news/2020/10/30/edition-68/,
 and the actual Rev News have a formatting which results in a very slim
left menu colomn. Failed :-( ... hints are appreciated. 

@chriscool: do you still plan to publish edition 69 tomorrow?

cc @jnareb @sivaraam

mjaix:
I had similar experiences when trying to enrich/augment IBM web pages
[...]

PR #471, opened 2020-12-02
css: improvements to "move the left margin"

felipec:
As @jnareb mentioned in issue #463, it should be possible to "move the left margin" so the navbar isn't squeezed.

However, conceptually what is actually being done is very different: the container wrapper is expanded to the full width of the window, then, the elements inside are justified to the center, this way the browser knows where to place the "left margin" (the left-most location where the navbar starts), but in order for the elements to not expand to the full width of the window, a max-width for the site container is necessary.

While on it, I adjusted the sizes of the navbar and the site containers, and removed some extra unnecessary padding.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -36,6 +36,7 @@ body {
     -webkit-flex: 1;
     -ms-flex: 1;
     flex: 1;
+    max-width: 660px;
 }
 
 .navbar {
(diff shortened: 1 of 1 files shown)


felipec:
This is what the changes looks like on the 68th edition: css-margin.

chriscool:
Ok, let's try this! Thanks @felipec !

Issue #468, opened 2020-11-27
Any comment about upcoming Git Rev News edition 70


chriscool:
A currently mostly empty draft is there:

https://github.com/git/git.github.io/blob/master/rev_news/drafts/edition-70.md

Feel free to comment in this issue, suggest topics, suggest persons to interview, or use the edit button (that looks like a pen) to edit and create a pull request with the changes you would like.

Let's try to publish this edition on Wednesday December 23rd 2020!

Thanks!

cc @jnareb @mjaix @sivaraam @gitster

chriscool:
Topics suggested for this edition:

from @felipec:

```
Shouldn't the [report](https://lore.kernel.org/git/742df4c2-2bc5-8a4b-8de1-cd5e48718398@redhat.com/) about git pull warning and the subsequent patch series trying to improve the situation be worthwhile to mention?

This BTW revives the [old discussion](https://lore.kernel.org/git/5363BB9F.40102@xiplink.com/), so it's interesting something that wasn't fixed in 2014 is still biting us today.
```

from @PhilipOakley:

```
Another thought for potential inclusions (given the `git pull` comment) could the the new and changed `config` variables (documentation & code) that have graduated to `next`/`main` so that these new / changed features and defaults become known. Probably something that can be deferred to the next Git Rev News
```

felipec:
Shouldn't the proposal to include a [reftable library](https://lore.kernel.org/git/pull.847.git.git.1600283416.gitgitgadget@gmail.com/) be worth mentioning?

chriscool:
@felipec thanks for the suggestion! Usually the main issue is that I don't have time to write about all the interesting things going on, and it doesn't happen often that someone else writes an article about something going on. But I think it's ok. No one really expects us to summarize everything.

chriscool:
Let's not forget to talk about the Outreachy interns who were accepted and started working recently. They have also posted to the mailing list about their blog.

chriscool:
Also in a short news I'd like to mention @felipec 's [New sharness/test vim syntax](https://lore.kernel.org/git/CAMP44s1D-Zp3KDS+Hi74a=Lkc7Nc_0qiEzQEF0Pmj=bD8i+=JQ@mail.gmail.com/).

felipec:
The [thread](https://lore.kernel.org/git/20201209065537.48802-1-felipe.contreras@gmail.com/) where a .vimrc loader is included in addition to the syntax file might be more newsworthy.

[...]

PR #474, opened 2020-12-25
rn-70: mention Felipe's Vim syntax file

sivaraam:
I've added a short news item for the sharness syntax mentioned by Christian in https://github.com/git/git.github.io/issues/468#issuecomment-741012684

I'm not sure if I placed it in the right place. So, feel free to shuffle/re-word it as necessary :)


chriscool:
Great! Thanks @sivaraam !

PR #478, opened 2021-02-19
SoC-2021-Ideas: wrap <format> with backquote

harry-hov:
The "\<format\>" in the project "Use ref-filter formats in `git cat-file`" is not visible [here](https://git.github.io/SoC-2021-Ideas/). 

This PR is to fix that issue.


chriscool:
Great! Thanks!

PR #479, opened 2021-02-21
SoC-2021-Ideas: ensure plain links are hyperlinked

sivaraam:
Plain links need to be wrapped in angle-brackets to be
hyperlinked in the rendered version of the page. So,
surround plain-links with angle-brackets to ensure
they're hyperlinked.


chriscool:
Thanks Kaartic!

PR #486, opened 2021-03-27
Point to information about writing a good commit message

sivaraam:
Newcomers applying for an internship program might not be
aware of the commit message guidelines of our community.

So, linking to it from the "General Microproject Information"
might be helpful.


chriscool:
Thanks @sivaraam !

PR #487, opened 2021-03-27
Ensure plain links are linkified

sivaraam:
Linkify plain links in "General Microproject Information" to make sure its easier for students to access those links.


chriscool:
LGTM! Thanks @sivaraam !

PR #494, opened 2021-04-07
Change default.html

snxx-lppxx:
Made the setting in the default file. If this change fixes the old one, I'll delete the old one.


chriscool:
Thanks @snxx-lppxx! Let's try this!

chriscool:
Thanks @sivaraam also for trying this and your comments on https://github.com/git/git.github.io/pull/493 !

PR #497, opened 2021-04-17
Fix icon path

### LINKED ISSUE #496: Favicon: missing
Change default.html #494
```
https://git.github.io/rev_news/rev_news/
https://git.github.io/rev_news/news_sources/
https://git.github.io/rev_news/archive/
```
Maybe create a separate html file?

snxx-lppxx:
Fix #496
----
Not sure. Check this, please.


_layouts/default.html
```diff
@@ -5,7 +5,7 @@
         <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
         <title>{{ page.title }}</title>
         <meta name="viewport" content="width=device-width">
-        <link rel="icon" type="image/x-icon" href="../images/logo.png"/>
+        <link rel="icon" type="image/x-icon" href="images/logo.png"/>
```
chriscool, line 8: It seems that @sivaraam suggested either `/images/logo.png` or just adding a link to it or a copy of it named `favicon.ico`. But this is doing something else. Is there a reason why you think Kaartic's suggestions will not work, or what you are doing is better?

snxx-lppxx: If I'm not mistaken, this is how it works:
```
_layouts
    L default.html
images
    L logo.png
```
And what I did then - it worked correctly, but I don't know why: `rev_news / rev_news.md` do not work. 
But here's what confused me:
https://github.com/git/git.github.io/issues/496#issuecomment-821791547

sivaraam: As far as I understand, paths in the layout are referenced relatively from the path **where the layout is used**. So, your relative reference worked properly for the index and about pages but not for Rev News pages.

That's why I suggested using an absolute path instead.

_layouts/default.html
```diff
@@ -5,7 +5,7 @@
         <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
         <title>{{ page.title }}</title>
         <meta name="viewport" content="width=device-width">
-        <link rel="icon" type="image/x-icon" href="../images/logo.png"/>
+        <link rel="icon" type="image/x-icon" href="images/icons/favicon.ico"/>
```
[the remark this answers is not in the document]
sivaraam: Alternatively, we could fix this `link` tag to make it work with the logo by using an absolute link as follows:

```suggestion
        <link rel="icon" type="image/x-icon" href="/images/logo.png"/>
```

sivaraam (changes requested):
The `favicon.ico` file should be placed at the ROOT of the repository. That way the favicon is automatically served to the browser upon request.

On some research though, it looks like using the `link` tag is the standardized way to serve website icons [[ref](https://en.wikipedia.org/wiki/Favicon#How_to_use)]. See suggestion about how to do that below.

snxx-lppxx:
Hello. Sorry for the lack of feedback.
<br/>
@sivaraam, check it out.

sivaraam (approved):
! Hello. Sorry for the lack of feedback.
No worries! 🙂 

! @sivaraam, check it out.
This looks good to me. Thanks for the fix @snxx-lppxx !

PR #501, opened 2021-05-02
rn-74: support: add a link to the mail that has the behaviour table

sivaraam:
While reading the support article, I felt that linking to Elijah's
e-mail that tabularizes the behaviour for the revert,
cherry-pick commands in various cases would be useful.
    
So, adding a link to it.


mjaix:
IMHO a helpful addition to illustrate the complexity.

chriscool:
Yeah, thanks both!

PR #504, opened 2021-05-26
rn-75: Add article "The top 1% of commit trailers"

Cogito:
Saw this fun little post from @felipec and thought it would make a nice article.

I tried to do a little summary and add some comments around trailers I thought were interesting, but please feel free to add/change/suggest others.

@felipec as it was your post happy to defer to you if there was anything in particular you wanted to say or highlight.

### DIFF
--- a/rev_news/drafts/edition-75.md
+++ b/rev_news/drafts/edition-75.md
@@ -38,6 +38,28 @@ This edition covers what happened during the month of April 2021.
   couldn't be selected! We hope to continue to see you in the
   community!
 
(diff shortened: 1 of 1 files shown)


felipec:
I would just add that it roughly follows [Zip's law](https://en.wikipedia.org/wiki/Zipf%27s_law); the 10th most popular line ("improved-by") is about 1/10 as popular as the 1st.

chriscool:
Thanks @Cogito and @felipec ! I plan to merge this before publishing this edition tonight, and to make the change @felipec suggests if you don't update this PR before.

Cogito:
Sounds good! Probably won’t get to update too soon so feel free to make changes as wanted.

PR #507, opened 2021-05-31
Add Mentoring-Program-Guide.md

chriscool:
Hopefully such a guide can help participants in a GSoC or Outreachy, and their mentors too.


chriscool:
No one has complained and we can always make further changes if this happens, so let's merge this.

PR #508, opened 2021-06-08
Update IRC link to Libera Chat

tfidfwastaken:
Libera Chat seems to be the go-to place for a lot of Git developers, and the last standup meeting happened there. Safe to say, new contributors are more likely to get their questions answered there.

Since there seems to be no way to directly link Libera's web chat to `#git-devel` directly, I have put a generic link over the text 'Libera Chat'. Users will have to manually type `#git-devel` in the webchat interface, but I suppose that should be obvious enough to not be worth mentioning.

### DIFF
--- a/Hacking-Git.md
+++ b/Hacking-Git.md
@@ -90,7 +90,7 @@ suggest improvements. Thanks!
 
 * [git-scm community page](https://git-scm.com/community)
 
-* [freenode webchat for the #git-devel IRC channel](https://webchat.freenode.net/#git-devel)
+* [web interface to #git-devel IRC channel on Libera Chat](https://web.libera.chat/#git-devel)
 
 * [#git-devel IRC channel archive](https://colabti.org/irclogger/irclogger_logs/git-devel)
 


Hacking-Git.md
```diff
@@ -90,7 +90,7 @@ suggest improvements. Thanks!
 
 * [git-scm community page](https://git-scm.com/community)
 
-* [freenode webchat for the #git-devel IRC channel](https://webchat.freenode.net/#git-devel)
+* #git-devel IRC channel on [Libera Chat](https://web.libera.chat)
```
sivaraam, line 93: How about this link: https://web.libera.chat/?#git-devel

tfidfwastaken:
Thanks @sivaraam. I fixed the link. I hope it's not too awkwardly phrased now.

Just doing a straight substitution of freenode webchat->Libera Chat webchat sounds very weird.

Another alternative I considered is "webchat for #git-devel on IRC" and drop the network name entirely, but decided against it, as it might cause confusion, for those who try to use their own client.

chriscool:
Great! Thanks @tfidfwastaken and @sivaraam !

PR #510, opened 2021-06-14
Mentoring-Program-Guide: make lines break

tfidfwastaken:
Add two spaces at the end of the numbered list, so that it does not all come on the same line. Makes it easier to scan, and formats it more closely to what I think the author intended.


chriscool:
Thanks @tfidfwastaken !

PR #513, opened 2021-06-17
README: discourage Git bug reporting

chriscool:
Let's try to avoid people using issues in this repo to discuss/report Git problems.

### DIFF
--- a/_includes/README.md
+++ b/_includes/README.md
@@ -1,12 +1,19 @@
-# About
+# About Git Developer Pages
 
-This homepage is maintained by editing files in the
+This is a website to help Git Developers.
+
+It is NOT a place to discuss Git issues. Please see
+[git-scm.org's community page](https://git-scm.com/community)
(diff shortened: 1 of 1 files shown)


peff:
We've had some success in git-scm repository with using templates. E.g., try https://github.com/git/git-scm.com/issues/new/choose, which is driven by the files in https://github.com/git/git-scm.com/tree/main/.github/ISSUE_TEMPLATE. It's not perfect, of course, but at least it puts the information in front of people right when they're trying to open the issue.

chriscool:
@peff Yeah, that's a good idea! Thanks for the links!

sivaraam (approved):
The changes here look good. But after seeing the issue template suggestion which seems like a much better solution to the problem. I wonder if we want to do this 🤔

chriscool:
@sivaraam let's merge this now and try to do the issue template suggestion soon. We can revisit this after the issue template has been improved.

PR #514, opened 2021-06-17
index: make this website definitive

chriscool:
Let's acknowledge that this website isn't an experiment anymore and that it contains information about Git Rev News and the Git project involvement in the GSoC and Outreachy mentoring programs.

### DIFF
--- a/index.md
+++ b/index.md
@@ -6,19 +6,23 @@ navbar: false
 
 # {{ page.title }}
 
-This is a tentative homepage for information on Git development. If you
+This is a website for information on Git development. If you
 stumbled into this by mistake, you may want:
 
   - <http://git-scm.com>, which has information on running
-    git and links to download the latest version
+    Git and links to download the latest version
 
   - <http://git.wiki.kernel.org>, the wiki that has historically
     contained developer information
 
 These pages are intended to collect information useful to Git
-developers, including collaborative editing of documents (i.e., it is an
-alternative to us having a wiki, but one that is edited entirely via git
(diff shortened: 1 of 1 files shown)


peff (approved):
Wow, it's quite a while since I wrote that "tentative" paragraph. I think it's long past time to get rid of it. :)

Thanks for all your work in making this site into a real resource.

index.md
```diff
@@ -6,19 +6,23 @@ navbar: false
...
+developers. This is also the web home of:
+
+  - the [Git Rev News newsletter](/rev_news/),
+  - the [involvement of the Git project in mentoring programs](/General-Application-Information/)
+    like [Outreachy](https://www.outreachy.org/) and the
+    [GSoC (Google Summer of Code)](https://summerofcode.withgoogle.com/)
 
-Note that this page is an experiment and a work in progress. It may go
-away at any time if it turns out not to be useful.
+These pages are intended to be edited collaboratively (i.e., it is an
```
sivaraam, line 26: I'm curious if we should briefly mention how the editing should be done. Or at least direct people to the "About" page where some information about that could be found.

sivaraam (commented):
This looks good to me. Just one comment.

index.md
chriscool: I would be ok with a redirect to the "About" page. We can do it on top of this later though, so I will just merge this for now.

chriscool:
Thanks @peff and @sivaraam for reviewing!

PR #526, opened 2021-10-16
rn-80: mention "git autofixup"

krobelus:
This is not exactly new, but I think it's really useful, and it has only
been mentioned on the Git list once.
That page also references two blog posts; one that goes as far as explaining
the heuristic used to assign fixup commits, and another one that introduces the
use case to beginners, see https://github.com/torbiak/git-autofixup#articles


chriscool:
Great, thanks!

PR #538, opened 2022-01-26
Some mentoring guide updates

sivaraam:
I happened to give the mentoring guide a read and noted some things that could be improved. I've done changes for those which are tracked via this PR. Do take a look and let me know your thoughts.

One thing I wanted to note. There's the following line in the guide:
> As we think it's best that you decide on your own the best time to update your blog and send notifications, you should then also decide when it's best to have a chat with your mentors. 

But there doesn't appear to be any mention about students deciding their best time to update blog and send notifications before that. Would it be a good idea to mention that before in the section about "Regular updates" ?

### DIFF
--- a/Mentoring-Program-Guide.md
+++ b/Mentoring-Program-Guide.md
@@ -38,15 +38,12 @@ You might want to check which blogging tools previous Outreachy
 interns and GSoC students used. Using a Git related tools is of course
 a good idea.
 
-For example using GitLab and GitLab pages:
-
-https://mirucam.gitlab.io/outreachy_blog/
-https://matheustavares.gitlab.io/
-https://gitlab.com/MatheusTavares/matheustavares.gitlab.io/
+- [Miriam Rubio's Outreachy intership experiences](https://mirucam.gitlab.io/outreachy_blog/)
+- [Matheus Tavares's Blog](https://matheustavares.gitlab.io/) [ [source](https://gitlab.com/MatheusTavares/matheustavares.gitlab.io/) ]
 
 Using GitHub:
 
-https://charvi-077.github.io/
+- [Charvi Mendiratta's Outreachy Internship Experiences](https://charvi-077.github.io/) [ [source](https://github.com/charvi-077/charvi-077.github.io) ]
 
 If you already have something setup using other tools though, even if
 it's not Git based, it's perfectly OK to use what you already have,
@@ -57,7 +54,7 @@ the Git community, including your mentors. Your blog is also something
 that can help you in your career by reflecting positively on your
 abilities and dedication. A good example is Matheus' GSoC blog posts:
 
-https://matheustavares.gitlab.io/tags/gsoc/
+[Matheus Tavares' GSoC posts](https://matheustavares.gitlab.io/tags/gsoc/)
 
 ## Regular updates
 
--- a/Mentoring-Program-Guide.md
+++ b/Mentoring-Program-Guide.md
@@ -64,20 +64,20 @@ planning to work on and how.
 
 The best way to update everyone about your work is probably to post on
 your blog. We understand though that you might not want to put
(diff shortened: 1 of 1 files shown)


chriscool:
> But there doesn't appear to be any mention about students deciding their best time to update blog and send notifications before that. Would it be a good idea to mention that before in the section about "Regular updates" ?

Yeah, it could be. I need to take a look at this. Maybe after reviewing this PR...

Mentoring-Program-Guide.md
```diff
@@ -341,3 +341,12 @@ A few things that you might want to do are:
 - Don't forget using [GSoC] or [Outreachy].
 
 - Don't forget using [RFC] or [WIP], if it's relevant.
+
+# Conclusion
+
+Hope you got an idea about how things usually go during the internship.
```
chriscool, line 347: Maybe "internship/GSoC/mentoring period" as "internship" is for Outreachy only. In general it was not easy when writing this and other such docs to make sure the terms would work for Outreachy, GSoC and possibly other future programs. Maybe somewhere we should have some kind of internal docs/dictionary listing the terms we should use.

Mentoring-Program-Guide.md
```diff
@@ -341,3 +341,12 @@ A few things that you might want to do are:
...
+As a reminder, don't be surprised if your mentors ask for things that
+are a bit different than what is described here. Typically your mentors
+will expect that you become more autonomous.
+
+Wishing you luck during the internship!
```
chriscool, line 352: Same thing about "internship".

BTW maybe we should already work on scripts detecting and possibly replacing some terms.

Mentoring-Program-Guide.md
```diff
@@ -341,3 +341,12 @@ A few things that you might want to do are:
 - Don't forget using [GSoC] or [Outreachy].
 
 - Don't forget using [RFC] or [WIP], if it's relevant.
+
+# Conclusion
+
+Hope you got an idea about how things usually go during the internship.
+As a reminder, don't be surprised if your mentors ask for things that
+are a bit different than what is described here. Typically your mentors
+will expect that you become more autonomous.
```
chriscool, line 350: I like it! Maybe:

"Typically your mentors will expect that you become more autonomous over time and will adjust things depending on how autonomous you already are".

Mentoring-Program-Guide.md
```diff
@@ -57,7 +54,7 @@ the Git community, including your mentors. Your blog is also something
 that can help you in your career by reflecting positively on your
 abilities and dedication. A good example is Matheus' GSoC blog posts:
 
-https://matheustavares.gitlab.io/tags/gsoc/
+[Matheus Tavares' GSoC posts](https://matheustavares.gitlab.io/tags/gsoc/)
```
chriscool, line 57: Nice, thanks!

chriscool:
Except the few nits in the conclusion, this looks great to me! Thanks!

Mentoring-Program-Guide.md
sivaraam: > Maybe somewhere we should have some kind of internal docs/dictionary listing the terms we should use.

Yeah. That would be good to have. I'll try and see if I could find the time to come up with this one.

Mentoring-Program-Guide.md
sivaraam: > BTW maybe we should already work on scripts detecting and possibly replacing some terms.

Yeah. If we have a glossary, this is definitely a good next step.

sivaraam:
> Except the few nits in the conclusion, this looks great to me! Thanks!

Thanks Christian! I've made changes based on your suggestions. Kindly take a look. I'll see if I could find time to look into that glossary for program generic terms that we could use.

chriscool:
Thanks again @sivaraam !

PR #539, opened 2022-01-26
Some miscellaneous changes

sivaraam:
Noticed a possible sentence rephrase to the General Application Information which I've done.

Also, ignored the vendor directory that gets generated on running `bundle install`.


chriscool:
This LGTM! Merged! Thanks!

PR #540, opened 2022-01-29
Add GSoC 2022 material and hide 2021 material

sivaraam:
The ideas document is mostly empty now. Ideas are yet to be gathered
from the community and added.

We also hide the SoC 2021 documents and Outreachy microproject documents
as they're stale now. The SoC 2021 org application isn't hidden as
we haven't applied for 2022 yet.

**Note**: I'm creating the ideas doc without any ideas to make it easy for others to open a PR, if necessary.

### DIFF
--- a/SoC-2021-Ideas.md
+++ b/SoC-2021-Ideas.md
@@ -1,6 +1,7 @@
 ---
 layout: default
 title: SoC 2021 Ideas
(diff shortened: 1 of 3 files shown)


chriscool:
LGTM, thanks!

PR #541, opened 2022-01-29
rn-83: spell out Git for Windows

sivaraam:
It's likely better to spell out Git for Windows as not everyone
might immediately understand what Git4Win refers to.


chriscool:
Great, thanks!

Issue #545, opened 2022-01-31
Any alternate method to view the latest drafts or PRs


PhilipOakley:
I tend to use the GitHub interface to propose items in the light reading section. 

The README has a multi-step Linux approach to viewing the drafts. Is there a way of directly viewing a draft proposed / saved changes during the edit (I'm mainly on Git-for-Windows)? e.g. some special version of the edit's link address

Also, could the README include details of how to get to the released rev_news editions, i.e. getting from [/_posts/2022-01-31-edition-83.markdown](https://github.com/git/git.github.io/blob/master/_posts/2022-01-31-edition-83.markdown) to https://git.github.io/rev_news/2022/01/31/edition-83/ ?

I could suggest some words along the lines of look to the `_posts` directory and then re-arrange the markdown filename to match the release.. 

Meanwhile the https://github.com/git/git.github.io/blob/master/rev_news/archive.md file has a bad link (404) to the "About" - which should it be - e.g. the top level README or https://git.github.io/about/ ?

chriscool:
Sorry for the slow answer. Honestly I don't know much about Jekyll and how it works even on Linux, and I hardly ever use it.

So I think it's OK if you (or anyone else actually) send PRs that haven't been tried with Jekyll. We will figure out something if there are discrepancies between what is expected and what's actually displayed on the site. Maybe we should say that in the README.

Thanks for you PR to update the link and for your interest in improving the site!

PR #547, opened 2022-02-01
archive.md: Update `About Rev News` link

PhilipOakley:
The current `About Rev News` link results in a 404.
Link direct to the git.github.io page.

Fixes one point in #545 

Signed-off-by: Philip Oakley philipoakley@iee.email

### DIFF
--- a/rev_news/archive.md
+++ b/rev_news/archive.md
@@ -5,7 +5,7 @@ title: Git Rev News Archive
 
 # Archive
 
-Here you can see all the previous editions. See [About Git Rev News](/rev_news/rev_news/) for ways you can subscribe.
+Here you can see all the previous editions. See [About Git Rev News](https://git.github.io/about/) for ways you can subscribe.
 
 <ul>
   {% for post in site.posts %}


mjaix (approved):

mjaix:
Thanks a lot for finding just another relative link!

PhilipOakley:
Thanks. I forgot to point out that the error I saw was when I navigated, on GitHub, **within** the repo to the `archive.md` file (i.e. here https://github.com/git/git.github.io/blob/master/rev_news/archive.md).

As it is a markdown file, GitHub renders it and showed the hyperlink. It was clicking on **that** hyperlink that failed, rather than me checking the gh pages at https://git.github.io/

In retrospect I'm not so sure that my change is the right one as I changed from (apparently) a relative link, to an absolute link. It maybe a relative link could still be used, but with some 'corrected' path. Also there are two "About" pages in the https://git.github.io/ pages, one of which is https://git.github.io/rev_news/rev_news/ 

I suspect most of the #545 comments are about that confusion between starting at the 'gh pages' vs starting at the repo' markdown.md pages.

sivaraam:
> Also there are two "About" pages in the https://git.github.io/ pages, one of which is https://git.github.io/rev_news/rev_news/

Precisely, I believe linking to https://git.github.io/rev_news/rev_news/ would be more appropriate as the link promises to give information about subscribing which https://git.github.io/about does not provide directly. The "About Git Developer Pages" does have a link to the "About Git Rev News" page but that's likely a roundabout way to get the information about subscribing to the newsletter.

>  It maybe a relative link could still be used, but with some 'corrected' path.

I couldn't immediately think of a some 'corrected' path that would work in both GitHub's Markdown rendering and the rendered site.  But I suppose, we don't expect too many people to read the pages of the site via GitHub. So, it might be much of a problem having links that don't work on GitHub. OTOH, we do seem to use absolute links a lot (in rev news posts, about page). So, it might be fine to have absolute links too.

I suppose for now, if there's one change to do now. It would be to change the absolute link in archive.md from https://git.github.io/about/ to https://git.github.io/rev_news/rev_news/ 🙂

PR #550, opened 2022-02-09
Soc-2022-Ideas: Sparse Index Integrations

derrickstolee:
As discussed on the mailing list, here is an idea for a student project.

cc: @vdye, @sivaraam

### DIFF
--- a/SoC-2022-Ideas.md
+++ b/SoC-2022-Ideas.md
@@ -17,5 +17,57 @@ about refactoring projects versus projects that implement new
 features" in the [general application information](https://git.github.io/General-Application-Information)
 page though.
 
-### Idea X
-_TBD: Ideas to be added based on discussion with community_
+### More Sparse Index Integrations
+
+The [spare index feature](https://github.blog/2021-11-10-make-your-monorepo-feel-small-with-gits-sparse-index/)
+accelerates Git commands when using
+[sparse-checkout in cone mode](https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/).
+This works by modifying the on-disk index file in a way that includes
+"sparse directory" entries instead of only file entries. This requires
+care when enabling the sparse index for different commands, as custom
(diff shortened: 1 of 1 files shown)


SoC-2022-Ideas.md
```diff
@@ -17,5 +17,56 @@ about refactoring projects versus projects that implement new
...
+* `git write-tree`
+* `git apply`
+* `git am`
+* `git checkout--worker` (for parallel checkout)
+* `git merge-index`
+* `git rerere`
+
+Project Size: Medium
+
+Co-Sponsors:
```
sivaraam, line 70: I suppose it would be better to re-phrase this as "Co-mentors" / "Possible mentors". Sponsors doesn't seem to fit well here as there's no financial aid required for GSoC.

SoC-2022-Ideas.md
```diff
@@ -17,5 +17,56 @@ about refactoring projects versus projects that implement new
...
+* `git worktree`
+* `git write-tree`
+* `git apply`
+* `git am`
+* `git checkout--worker` (for parallel checkout)
+* `git merge-index`
+* `git rerere`
+
+Project Size: Medium
+
```
sivaraam, line 69: We [usually mention](https://git.github.io/SoC-2021-Ideas/) the programming language experience required for each project. To be consistent with that, we could mention the same here. It could possibly be mentioned like:

Language: C, shell (bash)

sivaraam (commented):
Thanks @derrickstolee ! The idea seems well written. It looks fine for the most part. Just a few minor comments and I think we can merge this one.

SoC-2022-Ideas.md
derrickstolee: Right. "Mentor" is the word I should have used.

chriscool:
Yeah, I like it! And it LGTM! Thanks @derrickstolee! Thanks also @sivaraam for a good review and good suggestions!

PR #559, opened 2022-02-18
SoC-2022-Ideas: add Hariom as possible mentor

harry-hov:
As discussed on the mailing list, added me as a possible mentor in `Unify ref-filter formats with other pretty formats`.

cc: @chriscool @sivaraam


chriscool:
Great, thanks @harry-hov!

PR #560, opened 2022-02-18
add logo to svg

snxx-lppxx:
Hi, Changed the display options for the layout. Something like this :)


chriscool:
Thanks for your interest in improving these web pages!

Could you explain a bit in the commit message of each commit and perhaps in the PR description what is improved by the changes you propose? It's not easy for people not very familiar with web UIs to understand what the issues are in the first place and what your changes are doing.

snxx-lppxx:
Useless.

PR #564, opened 2022-03-01
Add article to Light Reading section

brunofbrito:
I published [this article](https://hackernoon.com/how-to-really-use-git-10-rules-to-make-git-more-useful) on HackerNoon this week. I think it fits nicely in the "Light Reading" section 🙂


chriscool:
Great, thanks @brunofbrito !

Please note that it would be a bit better to start the commit subject with for example "rn-85:" to tell us that it's for edition 85.

PR #565, opened 2022-03-04
README: fix rendering of e-mail links in About page

sivaraam:
The e-mail links in README render properly in GitHub's rendering of
the README file. In the deployed site they render as plain text, though.

Correct this by fixing the formatting of e-mail links such that they
render properly in the deployed site.


chriscool:
LGTM, thanks @sivaraam !

PR #566, opened 2022-03-04
SoC-2022-Ideas: linkify email links

sivaraam:
E-mails were wrapped in backticks which resulted in them rendering
as code snippets.

Tweak the e-mails so that they are rendered in a linkified manner.
This would help students to easily copy e-mails address (or) send
e-mails to potential mentors.

### Note
I'm creating this one as a PR as I'm kind of unsure of the reason the e-mails have been wrapped in codeblocks rather than being linkified. Does wrapping the e-mails in code block offer some kind of spam protection?

### DIFF
--- a/SoC-2022-Ideas.md
+++ b/SoC-2022-Ideas.md
@@ -72,8 +72,8 @@ Difficulty: Medium
 Languages: C, shell(bash)
 
 Possible mentors:
-* Derrick Stolee `<derrickstolee@github.com>`
-* Victoria Dye `<vdye@github.com>`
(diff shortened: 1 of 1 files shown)


chriscool:
I don't think it offers spam protection. I think it's just that we historically haven't cared much about links but just wanted them to stand out. I am OK with merging this, but I will wait a bit in case someone disagrees.

chriscool:
Hopefully we will copy the updated style next time, so this is still worth merging even if it's not very useful anymore for this GSoC.

Thanks @sivaraam for this and sorry for merging it so late!

PR #567, opened 2022-03-23
rn-85: Add Tower for Mac 8.1 release and one more "Light Reading" suggestion

brunofbrito:
We released the 8.1 update to Tower today. I also published a new blog post, so here are 2 more additions to this month's newsletter, if you agree! :)


chriscool:
Great, thanks @brunofbrito !

PR #575, opened 2022-04-30
rn-86: add git worktree article to light reading

bpugh:
Sorry it's a bit last minute and I totally understand if it's too late for this month, but I just came across this article that shows an interesting use case for `git worktree` that I wasn't familiar with.


sivaraam:
It's not too late for the edition yet. Thank you for your contribution!

chriscool:
Thanks @bpugh and @sivaraam !

PR #581, opened 2022-05-25
Update latest Gerrit releases

lucamilanesio:


chriscool:
Thanks @lucamilanesio !

PR #585, opened 2022-06-29
rn-88: include article about vimdiff mergetool driver change

sivaraam:
This article was mentioned in a mailing list thread. Junio hinted at including the article in Rev News in that thread. Ref: https://public-inbox.org/git/xmqqv8smqzg6.fsf@gitster.g/


chriscool:
Great! Thanks @sivaraam !

PR #593, opened 2022-07-22
rn-89: Add Light Reading suggestion

brunofbrito:


chriscool:
Thanks @brunofbrito !

PR #595, opened 2022-07-28
rn-89: minor copy-editing of gitster interview

gitster:


chriscool:
Great, thanks @gitster !

PR #601, opened 2022-09-29
Update the Git rev news with the Gerrit events

lucamilanesio:
Add the news on the Gerrit events happening in October / November and announce the return of the face-to-face Gerrit User Summit with also includes JGit talks and discussions.


chriscool:
Great, thanks @lucamilanesio !

Issue #602, opened 2022-10-01
Any comment about upcoming Git Rev News edition 92


chriscool:
A currently mostly empty draft is there:

https://github.com/git/git.github.io/blob/master/rev_news/drafts/edition-92.md

Feel free to comment in this issue, suggest topics, suggest persons to interview, or use the edit button (that looks like a pen) to edit and create a pull request with the changes you would like.

Let's try to publish this edition on Wednesday October 26th 2022!

Thanks!

cc @jnareb @mjaix @sivaraam @gitster

chriscool:
A writeup about GSoC 2022 projects would be nice in this edition, as we couldn't have one in the previous one.

jnareb:
It is a pity that there was no detailed writeup about the contents of **Git Merge 2022** in previous edition. I could only find slides for two presentations, but no blog posts:

- [Git Internals: A Database Perspective](https://stolee.dev/docs/git-merge-2022.pdf) by Derrick Stolee, covered by the series of articles mentioned in edition 90
- [Git at GitHub Scale](https://ttaylorr.com/presentations/git-merge-2022.pdf) by Taylor Blau

When Git Merge was in Europe, there were videos from the conference posted some time after it on YouTube. Not so now; the latest playlist I could find on YouTube is [from Git Merge 2019](https://www.youtube.com/playlist?list=PL0lo9MOBetEFqBue4vNcTEnkBjgIQU1Q3). It's a pity.

chriscool:
@jnareb maybe they didn't post the videos of the Git Merge 2020 because of the Covid that changed a lot of their plans, and maybe they will post those of the Git Merge 2022 after a few months as they did before 2020.

@ttaylorr do you know if GitHub plans to post the Git Merge 2022 videos? Also are some slides available?

ttaylorr:
> I could only find slides for two presentations, but no blog posts:

@derrickstolee wrote a handful of blog posts which comprised his talk:

- https://github.blog/2022-08-29-gits-database-internals-i-packed-object-store/
- https://github.blog/2022-08-30-gits-database-internals-ii-commit-history-queries/
- https://github.blog/2022-08-31-gits-database-internals-iii-file-history-queries/
- https://github.blog/2022-09-01-gits-database-internals-iv-distributed-synchronization/
- https://github.blog/2022-09-02-gits-database-internals-v-scalability/

and I wrote one that was published recently (and one that was published last year) which comprised my talk:

[...]

PR #611, opened 2022-10-27
Some more articles for edition #92

sivaraam:
Add articles suggested in the comments of edition #92: [ [comment 1](https://github.com/git/git.github.io/issues/602#issuecomment-1283919090) ] [ [comment 2](https://github.com/git/git.github.io/issues/602#issuecomment-1291012811) ].

I've linked to recent monorepo related blog that I could think of as relevant. Do suggest other ones in case they would be relevant.


jnareb:
Thanks @sivaraam

PR #624, opened 2023-01-24
SoC-2023-Ideas: remove finished sparse index integrations

vdye:
Both 'git grep' and 'git rm' were completed as part of GSoC last year. Update the list to reflect which commands are still available for the project.


chriscool:
Thanks @vdye ! Please let us know if you would like to mentor again this year.

PR #629, opened 2023-02-27
SoC 2023: update sparse index command list

vdye:
Remove 'rev-parse', since it was already integrated with sparse index last year [1]. Additionally, reorder the list of commands to reflect the relative difficulty of sparse index integrations, taking into account both the additional work of adding sparse-checkout compatibility (if needed) and what has been made easier as a result of other commands' integrations.

[1] https://lore.kernel.org/git/69efe637a18786b289db79971e9e49137306b57c.1651005800.git.gitgitgadget@gmail.com


chriscool:
Thanks @vdye !

PR #633, opened 2023-03-21
SoC-2022-Microprojects.md: clarify exit-code microproject

ttaylorr:
The original description is somewhat confusing, and has been the source of real-world confusion on the mailing list.

Let's clarify what we are trying to promote and avoid, as well as give some examples and background about our integration test suite.

(Note: I don't think we ever wrote any new microproject ideas for GSoC 2023, and ISTM that we are just reusing the ones from 2022. That's fine, and is the reason that I made these changes in the SoC-2022 file instead of creating a new one).


chriscool:
Thanks for clarifying this micro-project!

PR #640, opened 2023-04-16
Include development info and re-shuffle home/about pages

### LINKED ISSUE #610: Include development info in README?
Previously the repo's README used to be a link to [_includes/README.md](https://github.com/git/git.github.io/blob/master/_includes/README.md). It's now a normal file with some information about the site. It is nice for [the landing page](https://git.github.io/).

Though, now it lacks basic instructions that would help someone to know how to spin up the site for development locally. Would it be helpful if we include it in the repo's README in some way?

sivaraam:
The information to spin up the site locally got buried inside _includes/README in a recent change. It actually belongs to the README present in the root of the repo, too since that is what users see when viewing the repository. Hence, move that information there.

The "About" page has long been very identical to the home page. So, shuffle the content in home and about pages and unify them into the Home page itself. For historical purposes, make the about page redirect to the homepage.

Fixes #610

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,4 @@ gem 'github-pages', group: :jekyll_plugins
 
 # gem "webrick", "~> 1.7" # N.B. we may not want to fix this Gemfile to Ruby 3
 gem "webrick"
+gem 'jekyll-redirect-from'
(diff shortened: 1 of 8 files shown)


chriscool:
LGTM, thanks @sivaraam!

PR #662, opened 2023-09-20
rn-103: Add new Tower for Windows release

brunofbrito:


chriscool:
Thanks Bruno!

If possible next time, the commit should have the same title as the PR. Something like "rn-103: add new Tower for Windows release" would be great.

Issue #660, opened 2023-09-01
Any comment about upcoming Git Rev News edition 103


chriscool:
A currently mostly empty draft is there:

https://github.com/git/git.github.io/blob/master/rev_news/drafts/edition-103.md

Feel free to comment in this issue, suggest topics, suggest persons to interview, or use the edit button (that looks like a pen) to edit and create a pull request with the changes you would like.

Let's try to publish this edition on Wednesday September 27th 2023!

Thanks!

cc @jnareb @mjaix @sivaraam @gitster

chriscool:
Suggestion by @jnareb from previous edition issue:

[Fetching too many tags?](https://public-inbox.org/git/274ec1a2152b0fd53b35c1591f5177e0b0713430@rjp.ie/t/#u) - contains interesting info about how Git fetches from remote repository

jnareb:
Would this issue include conclusion of Google Summer of Code (GSoC 2023) program for Git, or is it too early?

sivaraam:
Thanks for the nudge, @jnareb. I think we're good to write about this in this edition.

If someone doesn't beat me to it, I'll look into opening a PR adding the GSoC conclusion :-)

chriscool:
@sivaraam thanks for writing about the GSoC conclusion!

@jnareb for your information the GSoC might not be finished for some projects as the timeline could be extended from 10 weeks up to 22 weeks, but it's finished for the 2 contributors who worked on Git and passed the final evaluation.

sivaraam:
I've opened #664 adding the conclusion of GSoC 2023.

chriscool:
Thanks @sivaraam, PR https://github.com/git/git.github.io/pull/664 is merged.

chriscool:
Draft email sent: https://lore.kernel.org/git/CAP8UFD1qDet8OPKYgk6dgibPT4PW=AeO7T3h=_vz5uYnEOzO3Q@mail.gmail.com/

I also emailed @ttaylorr to get a link to a read only version of the Git Contributor's Summit (which happened last Tuesday and Wednesday) notes.

chriscool:
> Suggestion by @jnareb from previous edition issue:
> 
> [Fetching too many tags?](https://public-inbox.org/git/274ec1a2152b0fd53b35c1591f5177e0b0713430@rjp.ie/t/#u) - contains interesting info about how Git fetches from remote repository

Thanks @jnareb for the suggestion! The article I wrote is about this.

[...]

PR #663, opened 2023-09-26
rn-103: Link to .gitkeep post

adamchainz:
Ref #660 for the upcoming edition.

If there’s an RSS reader that links are pulled from, my Git-related posts can be subscribed to at https://adamj.eu/tech/atom-git.xml .

I’ll be releasing my book “Boost Your Git DX” soon, probably before Rev News 104, so it would be nice to add it then.

Thanks 🙏

### DIFF
--- a/rev_news/drafts/edition-103.md
+++ b/rev_news/drafts/edition-103.md
@@ -40,6 +40,8 @@ __Various__
 
 __Light reading__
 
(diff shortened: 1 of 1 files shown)


chriscool:
Thanks @adamchainz!

Feel free to send a PR to add a link to your book when it is released. I will create the draft for edition 104 when edition 103 is published which should be around September 30 (next Saturday).

About subscribing to your Git-related posts, I will leave that to @jnareb who takes care of finding and organizing most of the links we put in the "Other News" section.

adamchainz:
Thank you for the merge and info! 🙏

PR #665, opened 2023-10-01
# rn-103: minor copy edit

stepnem:


chriscool:
Thanks for your fixes @stepnem !

PR #667, opened 2023-10-08
rn-104: Link to Boost Your Git DX book

adamchainz:


chriscool:
@adamchainz great, thanks!

PR #682, opened 2024-01-30
rn-107: Add new Tower for Windows release

brunofbrito:


chriscool:
Thanks @brunofbrito !

PR #683, opened 2024-01-30
rn-107: Add git config article

bpugh:
Just suggesting an article I recently wrote if you all think it's a good fit.


chriscool:
Thanks! I am Ok with this but I will let @jnareb decide as he manages this section.

jnareb:
Thanks @bpugh . I'll adjust it slightly in post.

bpugh:
Thanks @jnareb! 
By the way, I noticed your comment on the draft, and realized my post could have been more clear that I recommend setting the `rebase` option if you're specifying the flag every time and not necessarily that everyone should switch to pulling with rebase.
So I updated the post to add a quick clarification. I hope that ok.

Thanks again!

PR #684, opened 2024-01-31
rn-107.md: add more details and clarifications on rename detection

newren:


rev_news/drafts/edition-107.md
```diff
@@ -27,83 +27,119 @@ This edition covers what happened during the months of December 2023 and January
 
 ### Support
 
-* [Git Rename Detection Bug](https://public-inbox.org/git/LO6P265MB6736043BE8FB607DB671D21EFAAAA@LO6P265MB6736.GBRP265.PROD.OUTLOOK.COM/)
+* [Git Rename Detection Surprises](https://public-inbox.org/git/LO6P265MB6736043BE8FB607DB671D21EFAAAA@LO6P265MB6736.GBRP265.PROD.OUTLOOK.COM/)
```
chriscool, line 30: I prefer to keep the subject of the original email even if it's not accurate.

chriscool:
@newren thanks a lot for this suggestion which looks to me like a big improvement in general! Unfortunately, we usually have the subject of an article matching the subject of the email that started the thread. I know that the subject is not accurate in this case, but I think reader will realize it when reading the article. And we all know that all bug reports, often don't actually report real bugs.

chriscool:
So I would like to merge this but put back the original title. Does it work for you?

newren:
> So I would like to merge this but put back the original title. Does it work for you?

Sure, that's fine.  Do you want me to make the change or are you happy to make it?

chriscool:
I will make it, thanks!

chriscool:
Done in 3282d2607897af5093df699b2d3e42ce20be9a7f.

PR #693, opened 2024-02-24
SoC-2024: include the organization logo at a prominent place

sivaraam:
This is a suggestion by Google.

The change right now renders as follows locally.

image

## An alternative rendering

<details><summary>If we use simple markdown syntax for including an image, it renders as follows.</summary>
<p>

image

</p>
</details> 

I'm fine with either rendering. Feel free to share your thoughts. 🙂

### DIFF
--- a/SoC-2024-Ideas.md
+++ b/SoC-2024-Ideas.md
@@ -3,6 +3,8 @@ layout: default
 title: SoC 2024 Ideas
 ---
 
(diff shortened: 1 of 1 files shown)


chriscool:
Yeah, let's just do it for now, and then we can improve on it later, if we have some preference for a different rendering or other such changes.

sivaraam:
It renders a bit awkwardly on mobile. So, I suppose we need to consider tweaking it at less for small width screens. Let me see if something could be done about it 🙂

PR #694, opened 2024-02-25
SoC-2024: improve positioning of logo in small screens

sivaraam:
We could align the image to right only on bigger screens as it looks good only when the space on the left is enough for some decent amount of content to be shown. On smaller screens, we could just place the image on its one chunk of space (default behaviour when including images in Markdown) and let the content follow it.

Tweak the styles to accordingly for this.

<details><summary>Screen shot of page on big screen</summary>
<p>

image

</p>
</details> 

<details><summary>Screen shot of page on small screen</summary>
<p>

image

</p>
</details>

### DIFF
--- a/SoC-2024-Ideas.md
+++ b/SoC-2024-Ideas.md
@@ -3,7 +3,7 @@ layout: default
 title: SoC 2024 Ideas
 ---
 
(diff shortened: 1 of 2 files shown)


chriscool:
Looks great! Thanks @sivaraam !

PR #699, opened 2024-03-04
SoC-2024-Microprojects: add reference to help distinguish valid conversions

sivaraam:
Link to the helpful email by Eric that would help distinguish valid and invalid conversions of "test -e"


chriscool:
Thanks @sivaraam !

PR #703, opened 2024-03-28
SoC-2024: clearly mention that the Git project can accept upto 3 projects

sivaraam:
Christian has clarified that he could mentor one project alone and another along with Sivaraam. So, adjust the project limit to clearly clarify that we could accept up to 3 projects.


chriscool:
Thanks Kaartic!

PR #707, opened 2024-04-30
Fix typo in name

LemmingAvalanche:


chriscool:
Thanks @LemmingAvalanche for the fix!

PR #708, opened 2024-05-01
rn-110: minor copy edit

stepnem:


chriscool:
Thanks @stepnem for the great fixes!

PR #715, opened 2024-06-07
Fixed Side Navigation Bar for Seamless Page Navigation

OmkarSonawane1230:
This pull request addresses the issue of the side navigation bar disappearing when scrolling down the page. By making the my content height match window height and overflow of scroll, users can now access menu options without needing to scroll back to the top of the page

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -36,6 +36,8 @@ body {
     -webkit-flex: 1;
     -ms-flex: 1;
     flex: 1;
+    max-height: 100svh;
+    overflow-y: scroll;
     max-width: 700px;
 }
 
--- a/css/main.css
+++ b/css/main.css
@@ -36,13 +36,13 @@ body {
     -webkit-flex: 1;
     -ms-flex: 1;
     flex: 1;
-    max-height: 100svh;
-    overflow-y: scroll;
     max-width: 700px;
 }
 
 .navbar {
   float: left;
+  position: sticky;
+  top: 0;
   text-align: left;
   width: 15em;
   padding: 2em;
@@ -66,6 +66,8 @@ body {
   display: -ms-flexbox;
   display: flex;
   margin: 0 auto;
+  max-height: 100vh;
+  overflow-y: scroll;
   justify-content: center;
 }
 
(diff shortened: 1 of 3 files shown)


chriscool:
@sivaraam could you take a look at this?

css/main.css
```diff
@@ -36,6 +36,8 @@ body {
     -webkit-flex: 1;
     -ms-flex: 1;
     flex: 1;
+    max-height: 100svh;
+    overflow-y: scroll;
```
sivaraam, line 40: Rather than making the a fixed-height for the whole page which might hurt accessibility a bit, could you try to make the sidebar sticky? 🤔

sivaraam (commented):
Thank you for the change. I have a suggestion. Kindly check the same.

css/main.css
OmkarSonawane1230: Okay I will do it that way

OmkarSonawane1230:
@sivaraam I change the navbar position to sticky, but in order to make it work I have to make the `#wrapper` height to window height and overflow scroll, now the position sticky will work when it is scrolled.

OmkarSonawane1230:
@sivaraam 

**I want to share some problems in the site**

- The `.site` container change its width according to its content, which create a **horizontal scrollbar** in some cases.
- Inconsistencies in the **font sizes** used in `.site`
- It is difficult to **determine which menu item was selected after clicking on it**. This lack of visual feedback made it challenging to navigate through the site.
- The `navbar` shrinks when there is not enough space to fit the `site` which make it unreadable.

**I have a solution for it**

I think we should change the **design** for side navigation, I design a layout to **solve the problem**:

new design for git github io

**_This design not what I am building it is just for explaining my point, the actual design will be different and can be changed further._**

In this layout the side navbar will hide on left side for small screen size, and it can be access by hamburger menu or side swap. Here it can will be easy to determine which menu is selected by the arrow in front of the menu item. there will be no horizontal scroll bar because we are using more horizontal space than before.

chriscool:
@OmkarSonawane1230 This design looks good to me, but it should be discussed in a separate issue. Could you open a new issue with the above description? Thanks.

PR #716, opened 2024-06-11
Enhanced Side Navbar Functionality 

OmkarSonawane1230:
<h3>Key changes include:</h3>

1. **Improved Side Navbar:** Remove the bullets, doesn't shrink when open some tabs and make the selected tab item bold.
2. **Elimination of Horizontal Scrollbars:** Horizontal scrollbars are now hidden in specific tabs
3. **Tab Indicator:** Implemented a visual indicator in the navbar to clearly identify the currently opened tab.

**Before**

Screenshot 2024-05-18 at 7 34 36 PM

**After**

Screenshot 2024-05-18 at 7 34 36 PM

<h4>**In smaller screen size**</h4>

| Before | After |
|--------|-------|
| Before Image | After Image |

### DIFF
--- a/script/script.js
+++ b/script/script.js
@@ -1,0 +1,15 @@
+const NAVBAR_LI_ELEMENTS = document.getElementById('navbar').firstElementChild.children;    // Access all the <li> in side navigation
+let flag = false;
+
+
+// rotate the arrow of <li> which matches the SITE_URL
+Array.from(NAVBAR_LI_ELEMENTS).forEach((liElement) => {
+    if (liElement.firstElementChild.href == location.href) {
+        liElement.classList.add('active');
+        flag = true;
+        return;
+    }
+})
+
+// if non of other is active the the Home tab is active
+if (!flag) NAVBAR_LI_ELEMENTS[0].classList.add('active');
--- a/script/script.js
+++ b/script/script.js
@@ -1,15 +1,9 @@
 const NAVBAR_LI_ELEMENTS = document.getElementById('navbar').firstElementChild.children;    // Access all the <li> in side navigation
-let flag = false;
-
 
 // check for a match with the <li> element and SITE_URL, then make it active
 Array.from(NAVBAR_LI_ELEMENTS).forEach((liElement) => {
     if (liElement.firstElementChild.href == location.href) {
         liElement.classList.add('active');
-        flag = true;
         return;
     }
 })
-
-// if no other tab is active, then the Home tab becomes active
-if (!flag) NAVBAR_LI_ELEMENTS[0].classList.add('active');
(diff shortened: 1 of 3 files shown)


script/script.js
```diff
@@ -0,0 +1,15 @@
...
+// rotate the arrow of <li> which matches the SITE_URL
+Array.from(NAVBAR_LI_ELEMENTS).forEach((liElement) => {
+    if (liElement.firstElementChild.href == location.href) {
+        liElement.classList.add('active');
+        flag = true;
+        return;
+    }
+})
+
+// if non of other is active the the Home tab is active
```
chriscool, line 14: Do you mean "if no other tab is active, then the Home tab becomes active"?

chriscool:
Except my question about a code comment, it LGTM. I'd still like @sivaraam to take a look though. Thanks!

script/script.js
OmkarSonawane1230: Yes, I accidentally misspell it.

OmkarSonawane1230:
@sivaraam if we don't set the main `wrapper` height to 100% then we can't apply the sticky property to the `navbar` because it is inside the `wrapper`, which make it move upward as page is scrolled, because the scroll is apply to the `html` it self. if we set the `wrapper` height 100%, then we can apply position sticky to `navbar` which will work as expected.

sivaraam:
@OmkarSonawane1230 I actually gave it a shot already. Try checking out the code in the following branch: https://github.com/git/git.github.io/tree/sticky-without-max-height

You will see the sticky achieved without setting the `max-height` for the whole wrapper.

OmkarSonawane1230:
@sivaraam Thank you for your feedback. I have addressed the issues you mentioned in the code review.

I followed your suggestion to refactor the `max-height` and `sticky position` problem. I was using safari so the `position: sticky` was not working on my side. but now it is working by adding `position: -webkit-sticky`.

Thanks again for your guidance on this.

sivaraam:
The changes now look good for the most part. There's just this one issue I noticed which doesn't exist in the current webpage. On smaller screens there's some additional space on the right which is resulting in an unnecessary horizontal scrollbar even though there isn't any content displayed there. Could you check the cause of this?

Here's a screen recording about the issue:

https://github.com/git/git.github.io/assets/12448084/95c39bd2-9893-41e1-8a3c-99be830f1d0c

OmkarSonawane1230:
@sivaraam the issue was caused navbar.

The `navbar` was causing the horizontal scrollbar because of `width: 100%`, so I changed it to `auto` now it works fine and the horizontal scroll is not visible any more.

sivaraam:
Looks great. Thanks, @OmkarSonawane1230 !

chriscool:
Thanks @OmkarSonawane1230  and @sivaraam !

PR #723, opened 2024-08-01
rn-113: minor copy edit, take two

stepnem:


chriscool:
Looks great to me, thanks @stepnem !

PR #725, opened 2024-09-01
rn-114: minor copy edit

stepnem:


chriscool:
Thanks @stepnem for the great fixes, and thanks @mjaix for reviewing and merging them!

PR #731, opened 2024-10-01
rn-115: minor copy edit

stepnem:


chriscool:
LGTM, so merged, thanks @stepnem !

PR #733, opened 2024-10-02
Microproject-Information: the review interaction is a cooperative process

gitster:
I admit that it was I who wrote the original of this section [*] but the use of verb "defend" here gives a wrong impression that it is a lot more adversarial exchange than it really is.  It is not like your opponents are trying to shoot down your proposal, against which you need to successfully defend.  It is more like the reviewers and the contributor together try to figure out a good solution to what the problem the proposal tries to solve.

And it takes _understanding_ from all sides what others are bringing to the discussion.  A better verb to use is "explain" to give others visibility into what you have in this paragraph.

[Reference]
 * https://lore.kernel.org/git/xmqq8usx4pvh.fsf@gitster.dls.corp.google.com/


chriscool:
Great, thanks @gitster !

PR #735, opened 2024-10-15
Added Git For Each Ref article link

bkuhlmann:
## Overview

This article details how to use `git for-each-ref` in new and interesting ways (including the new `is-base` field).

## Details

I published this article earlier today so thought it might be of interest to folks here too.

### DIFF
--- a/rev_news/drafts/edition-116.md
+++ b/rev_news/drafts/edition-116.md
@@ -40,6 +40,9 @@ __Various__
 
 __Light reading__
 
++ [Git For Each Ref](https://alchemists.io/articles/git_for_each_ref) by Brooke Kuhlmann. Learn how
+  to use this command to make use of references for information dumping, statistcs, and much more. Included in this article is use of the new `is-base` field name recently added in link:https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.47.0.txt[Git 2.47.0].
+
 <!---
 __Easy watching__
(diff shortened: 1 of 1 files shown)


LemmingAvalanche:
I skimmed it and it looks like an interesting edition to the newsletter. This is a command that probably needs more exposure.

Thanks.

rev_news/drafts/edition-116.md
```diff
@@ -40,9 +40,8 @@ __Various__
 
 __Light reading__
 
-<!---
-__Easy watching__
--->
```
chriscool, line 45: Please don't remove such commented sections. They are commented because we might want to uncomment them if we have relevant content (here videos) to put in them, and it's still possible at this point that we will have such content.

chriscool:
@bkuhlmann thanks for the suggestion and @LemmingAvalanche  thanks for your opinion on it. I will let @jnareb decide on this though as he is responble for the link.

rev_news/drafts/edition-116.md
bkuhlmann: Ah, apologies. Fixed. Was too hasty and accidentally deleted that section.

chriscool:
@bkuhlmann I had to fix the link to Git 2.47.0 Release Notes in 36edbf521bc5d8da9a4504885c1c3092b74ae1a8.

PR #736, opened 2024-11-01
rn-116: minor copy edit

stepnem:
I'm also overriding some previous edits, so perhaps some
explanation is in order:

1. 079a323e4a75 changed some 'open-source' to 'open source',
but not all and not consistently.  Standard English uses
dashed forms for adjectives and non-dashed forms for nouns;
follow that rule.

2. 079a323e4a75 changed "people's" to "peoples'", which is
not correct here (it would imply a number of ethnic groups
is meant, but that's not the case); restore the original.

(As an aside, personally I tend to leave the interviews
alone beyond obvious typos, as I consider the language
idiosyncrasies part of the glimpse into an interviewee's
personality.)

### DIFF
--- a/rev_news/drafts/edition-116.md
+++ b/rev_news/drafts/edition-116.md
@@ -69,7 +69,7 @@ This edition covers what happened during the months of September and October 202
   killed using the Control-C keys on a terminal.
 
   The exit code associated with a signal is '128 + the signal number',
-  for example as the Control-C keys send a `SIGINT` signal, which signal
+  for example as the Control-C keys send a `SIGINT` signal, whose signal
   number is 2, processes killed this way should exit with code '128 + 2',
   so 130.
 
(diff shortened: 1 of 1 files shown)


mjaix:
Thanks for your fixes! They look good to me, especially the people's one.
Regarding open source, I was also not completely happy with my changes.
From a strict English point-of-view, the adjective is written with a dash,
so open-source software is an example following that rule.
With community and ecosystem, however, I am not 100% sure whether "open source"
is an adjective or if we are talking of a compound noun - 
this is a detail where my German background might lead me the wrong way.

mjaix:
In addition to resolving the merge conflict, I deliberately added "the new" before the platform policy document link.

chriscool:
Thanks @stepnem and @mjaix for your work on these fixes!

Should I merge now? Or @mjaix do you want to merge this?

chriscool:
@mjaix, oh it looks like you just merged it. Thanks!

chriscool:
It's closed now actually. Maybe it was a spurious GitHub bug?

stepnem:
> From a strict English point-of-view, the adjective is
> written with a dash, so open-source software is an example
> following that rule.  With community and ecosystem,
> however, I am not 100% sure whether "open source" is an
> adjective or if we are talking of a compound noun - this
> is a detail where my German background might lead me the
> wrong way.

I don't see any difference (from a grammar or orthography
perspective) between "open-source X" with X being
"software", "community", "ecosystem" or another noun.

That said, the distinction is often simply not being made,
esp. in American usage, in that even the adjectives are
written without dashes, same as nouns.  This is also
regularly the case in Git Rev News, and I usually leave
those alone (we're not publishing a book here; I mostly just
try to fix things that I find actually distracting or
influencing reading comprehension).  This time I fixed your
edits because they were (in part) changing the (correct)
original into a less consistent (or even incorrect) state.

Thanks!

PR #746, opened 2025-01-31
rn-119: minor copy edit

stepnem:
(Just a couple nits I see in the current draft;
unfortunately I'm going to be offline during the weekend, so
if you publish tomorrow as planned I'm unlikely to be able
to provide any further input this time. I'll have another
look when I come back. :-)


chriscool:
Thanks @stepnem ! No worries about being offline. I will publish this edition, but further changes can still be made after that. @mjaix  is also offline, but might make changes after publication too.

PR #752, opened 2025-02-28
rn-120: mention Git's selection in GSoC

sivaraam:


chriscool:
Great, thanks Kaartic!

PR #760, opened 2025-03-24
microprojects: add notes for the 2025 microproject list

KarthikNayak:
From an ongoing discussion on the mailing list [1], it would be nice to add some notes to the 'Modernize a test script' microproject. Let's do that.

[1]: https://lore.kernel.org/all/CAPig+cRm+sc+Rk-4SuQ5CrPeZLG2Nzz9B7+6OZxCq7tV5mzmBA@mail.gmail.com/

### DIFF
--- a/SoC-2025-Microprojects.md
+++ b/SoC-2025-Microprojects.md
@@ -206,3 +206,8 @@ There should be only one kind of change per commit. For example if one
 of your commits indents test bodies with TABs, instead of spaces, then
 this should be the only kind of change in this commit.
 
+#### Notes
+- only work on t/t????-*.sh scripts.
+- pick just one script (so as to avoid exhausting the pool for other candidates).
+- only convert `test -[def]` instances which semantically are assertions
+  (i.e. used as part of a &&-chain).


chriscool:
Thanks @KarthikNayak !

gitster (commented):
This item is about "Modernize a test script", so your Notes #1 (only work on t/t????-*.sh) is trivially correct, #2 makes sense, BUT listing #3 together with them does not make any sense. The quoted message from Eric talks many things and only one bullet item among 17 bullets is about "test -[def]".

KarthikNayak:
> This item is about "Modernize a test script", so your Notes #1 (only work on t/t????-*.sh) is trivially correct, #2 makes sense, BUT listing #3 together with them does not make any sense. The quoted message from Eric talks many things and only one bullet item among 17 bullets is about "test -[def]".

Okay. I get what you're saying, it would make sense to clarify which point we're referring to. I thought that inference could be drawn, but better to make it explicit.

KarthikNayak:
@chriscool @gitster I have raised https://github.com/git/git.github.io/pull/761 as a follow up

chriscool:
Thanks @KarthikNayak and @gitster ! I have just merged #761.

PR #761, opened 2025-03-25
microprojects: clarify last note in modernization test scripts

KarthikNayak:
The last note in the modernization test scripts is a little vague. Clarify it to be more clear about which modernization point it is referring to from the ones in the quoted email.


chriscool:
Thanks @KarthikNayak and @gitster for improving this!

PR #769, opened 2025-04-07
Rename references from *.txt to *.adoc in documentation paths.

ayu-ch:
Since the documentation in git is changed from *.txt to *.adoc format. Update references to reflect that change.


ayu-ch:
Do we also change the references in previous editions of Rev News?

chriscool:
Thanks @ayu-ch !

About previous Git Rev News editions, yeah, I think it makes sense to change them in a separate PR. Thanks.

ayu-ch:
I've made a PR for Git Rev News: #770.

PR #770, opened 2025-04-07
Rename references from *.txt to *.adoc in Rev News editions.

ayu-ch:
Since the documentation in git is changed from *.txt to *.adoc format. Update references in previous editions to reflect that change.


chriscool:
Thanks @ayu-ch !

PR #771, opened 2025-04-11
Add guidelines for using AI tools

chriscool:
There have been discussions both between Git GSoC mentors among themselves and between all the GSoC mentors on the Google Summer of Code Mentors List (`google-summer-of-code-mentors-list@googlegroups.com`) about applicants using AI more and more. It seems related to a big increase in the number of spam proposals that Git and some other project have been receiving recently.

Let's try to address this issue by providing guidelines for using AI tools to mentoring program candidates.

The Conservancy also recently organized talks about the use of AI tools to develop open source software. I have tried to take this into account especially by adding a "Legal issues" section in the guidelines.

### DIFF
--- a/General-Application-Information.md
+++ b/General-Application-Information.md
@@ -9,7 +9,9 @@ programs can get information about what the Git project would like to
 see in an application.
 
 *Please read this page completely before focusing on a project or a
- microproject ideas, or microproject general information.*
+ microproject ideas, or microproject general information.
+ Specifically, also read the "AI guidelines" section to know our
+ stance regarding the usage of AI tools.*
 
 
 ## Microproject (required)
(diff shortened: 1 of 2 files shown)


General-Application-Information.md
```diff
@@ -130,6 +130,78 @@ are interested in, it is a good idea to:
...
+says basically that authors of a patch should be sure about the origin
+of its content.
+
+As we cannot usually be sure how AIs have been trained and that they
+are not just repeating proprietary existing code or docuementation
+they saw during their training, we cannot accept much AI generated
+code or documentation in general. Now, if it's only a few lines to fix
+a bug or to implement a common pattern or summarize something, and if
+that looks specific enough to a current concrete problem a developer
+is working on that might be OK.
```
KarthikNayak, line 156: I would drop this, mostly because we clearly state that we're unsure of how the legality works out. Perhaps we can add something like:

> Only use such tools at your own discretion.

General-Application-Information.md
```diff
@@ -130,6 +130,78 @@ are interested in, it is a good idea to:
...
+Developers would often likely get better results, learn more and have
+a better overall experience by asking AIs to only explain things, and
+guide them step by step towards producing a solution by themselves,
+rather than by asking for a full solution that they would then mostly
+copy-paste.
+
+They can also use AIs to help with debugging, or with checking for
+obvious mistakes, things that can be improved, things that don't match
+our style, guidelines or our feedback, before sending it to us.
+
```
KarthikNayak, line 204: Should we also add a point, that applications which are AI generated and not following our guidelines, are very like to be dropped?

KarthikNayak:
Thanks @chriscool this looks good, some small nits from me. Feel free to ignore! :)

General-Application-Information.md
chriscool: I think it's a general rule about copyrighted work that if there are only a few ways, or a common way, to do something, then those cannot be copyrighted (see https://en.wikipedia.org/wiki/Sc%C3%A8nes_%C3%A0_faire).

I don't want to put that Wikipedia link in the doc, as I don't think it's worth distracting people with this.

General-Application-Information.md
chriscool: Yeah, right, I have just force pushed a new commit that adds the following:

```
For another, unfortunately common, example, it's unacceptable to send
us an application that has obviously been AI generated and doesn't
follow our guidelines or the feedback we previously gave to other
applicants. Those applications will be dropped.

```

General-Application-Information.md
```diff
@@ -130,6 +130,83 @@ are interested in, it is a good idea to:
...
+match our guidelines or our feedback.
+
+Yeah, AIs still often hallucinate or just produce bad outputs, even
+when you point out their mistakes.
+
+It's unacceptable to send us something that is obvious AI slop, or
+that sounds overly formal or bloated, or that looks good on the
+surface but makes no sense, or that senders don’t understand or cannot
+explain. It just wastes our time and we cannot accept it.
+
```
shejialuo, line 188: Someone has sent a lot of AI-generated patches to the mailing list and even to my personal email without sending to the mailing list.

So, instead of applications, should we also mention that we don't allow this for microproject. And should we add a rule if the students violate the rule, we won't consider them.

chriscool: @shejialuo, yeah right, that might help too. I have added the following:

```
In general, it's unacceptable to send AI-generated patches or messages
as-is to the mailing list or to mentors' or developers' personnal
email addresses. We won't consider candidates doing that.
```

I have also added a section called "Blindly using AI is far worse than not applying" and tried to improve other things in the new version I just sent.

Thanks!

sivaraam (commented):
Overall, this seems like a good addition. Just a few things that I think could be corrected.

General-Application-Information.md
```diff
@@ -130,6 +130,110 @@ are interested in, it is a good idea to:
 used for searching the mailing list and linking to previous
 discussions.)
 
+## AI guidelines
```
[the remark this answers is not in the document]
chriscool: Yeah, right, I have added this. Thanks.

General-Application-Information.md
```diff
@@ -130,6 +130,110 @@ are interested in, it is a good idea to:
...
+more and more popular, but that's not a reason to blindly use them,
+especially when working on the Git project.
+
+### Legal issues
+
+In general, it's not clear if AI generated code or documentation are
+acceptable in the Git project. Authors contributing to Git are
+required to "sign off" the patches they contribute, and to agree with
+the Developer’s Certificate of Origin, also called DCO, (see
+https://git-scm.com/docs/SubmittingPatches#sign-off), while the DCO
```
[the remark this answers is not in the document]
chriscool: Done. Thanks @sivaraam!

sivaraam:
Sure, Christian. There has been one thing that's been bugging me regarding this for a bit. The guidelines freely cover the proposal related points and code related points across sections. I wonder if it would make sense to restructure the sections a bit to distinguish our thoughts regarding both more explicitly and have general points separately. 🤔

chriscool:
@sivaraam I am not sure we can make a very significant distinction between code and proposal. In particular, as part of code contributions, writing commit messages and updating documentation share some aspects with writing a proposal.

So I wouldn't be against it if there is a good way to do it, but I don't see a good one right now. If you do, feel free to send a separate PR when this one is merged.

PR #772, opened 2025-04-14
Add CI instructions for GitLab

KarthikNayak:
The CI instructions are only present for GitHub. Since GitLab CI is also supported, let's add some documentation there for GitLab.


chriscool:
Thanks @KarthikNayak !

PR #773, opened 2025-04-29
rn-122: Add Light Reading + Tower for Mac release

brunofbrito:


chriscool:
Thanks @mjaix and @brunofbrito !

PR #781, opened 2025-06-24
rn-124: Add new Tower releases + Light Reading

brunofbrito:


chriscool:
@gitster, "git-tower" is in the web site's name. It looks like the product is actually called "Tower".

Also I am open to discussing potential trademark issues, but I don't think Git Rev News issues are the right place for that. We have other channels for that.

So I am going to merge this.

chriscool:
@jnareb feel free to edit the Light Reading part of this.

PR #785, opened 2025-07-02
minor copy edit

stepnem:


chriscool:
Thanks @stenem!

@jnareb the new link to the Combining branches zine looks good to me, but you might want to check it.

PR #787, opened 2025-07-13
Hacking-Git: remove review club from the list of resources

sivaraam:
Ref: https://lore.kernel.org/git/CAFySSZBgG3QyqEXSs3t4LzGYw1VJrHPx4gomT0zSsWcvQLezAQ@mail.gmail.com/


chriscool:
A link to the meeting notes could still be interesting for those who would like to see what happened, but I guess they could find it from the history of this page or from the mailing list archive. So I am fine with removing it too.

Thanks @sivaraam!

PR #789, opened 2025-07-24
Link git-phoenix in Hacking-Git.md

yaitskov:


yaitskov:
How about dedicated section for recovery tools next to debugging eg?

The link in monthly digest is not possible discover by hand.

chriscool:
The Hacking-Git page is specifically for developers tools or resources, especially those that could be useful to people starting to work on Git. So we have to draw the line somewhere. Creating a section for recovery tools could open the door to a lot of requests for other tools or resources or articles to be added.

There is a reason why we have the following at the top of the page:

"The goal is also not to list all the articles about Git or its internals. There are a lot of good resources, including free [books](http://git-scm.com/book/en/v2/), about that elsewhere."

It says "articles" but it's true for tools too.

We used to have a Wiki (https://archive.kernel.org/oldwiki/git.wiki.kernel.org/index.php/Main_Page.html) with the following near the top:

"[Git Tools](https://archive.kernel.org/oldwiki/git.wiki.kernel.org/index.php/Interfaces,_frontends,_and_tools.html) – Note: Have you written one? Add it here and announce it on the [list](https://archive.kernel.org/oldwiki/git.wiki.kernel.org/index.php/GitCommunity.html)!"

but it stopped being maintained, and now we just promote Git related tools, articles and resources in our Git Rev News newsletter.

You say that there "must be some registry with categories available for users for exploration", but we tried to do that and no one was willing to maintain the wiki for a long time, so the tools and articles listed became soon obsolete.

Anyway these days most people use search engines to look for tools or articles or resources, and we believe that listing some resources we like in our news letter is enough to give them a significant boost so that they will more likely be shown in the top search results.

chriscool:
By the way https://github.com/git/git.github.io/pull/790 was merged by @jnareb, so we will indeed talk about git-phoenix in the next edition of the Git Rev News.

PR #796, opened 2025-09-02
rn-126: minor copy edit

stepnem:
Two changes which might warrant explanation:

1. 'misguided thoughts' is the title of Sean Gillespie's
blog, but its lowercase spelling gives opportunity for the
sentence to be misunderstood as editorial
commentary (unlikely, granted, but gives the reader
unnecessary pause); another option I considered was adding
quotes or linkification, but saying just "on his blog" seems
perfectly sufficient and clear.

2. Add quotes around a part of the git-tools description to
indicate that "help you make crazy stuff in a very unsafe
way" is a direct quote of the tool author's description.
Otherwise I feel that this, too, will give the reader pause,
as it provides almost no information, in contrast with what
we usually try to achieve with descriptions.

### DIFF
--- a/rev_news/drafts/edition-126.md
+++ b/rev_news/drafts/edition-126.md
@@ -257,7 +257,7 @@ welcome your thoughts and feedback!_
 
 * **Have you thought about mentoring new GSoC / Outreachy students?**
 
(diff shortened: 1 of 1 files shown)


chriscool:
Looks great to me, thanks @stepnem!

PR #799, opened 2025-09-12
gsoc-participants: add partial 2025 participant information

sivaraam:
Creating this draft MR so that we don't miss this. We'll need to wait until program is actually complete to include archive links.


chriscool:
@sivaraam, thanks! Yeah, let's merge this when we have all the archive links.

chriscool:
Looks great to me, thanks @sivaraam !

PR #815, opened 2026-01-01
Git Rev News: Garden and Git Cola end of year releases

davvid:


davvid:
These are already handled by `get_releasees.py` but I released these just before the deadline so I wanted to make sure they weren't missed. Thanks again!

PR #668, opened 2023-10-15
fixed width that's lesser than the content width #472

Ananyasingh2002:
The website's responsiveness was enhanced with "flex-wrap: wrap," and the layout adjustment at "@media (max-width: 993px)" ensures optimal display on various devices, creating a fully functional and user-friendly experience. [Ensure code-blocks take-up fixed width that's lesser than the content width #472](https://github.com/git/git.github.io/issues/472)

PR #816, opened 2026-01-01
Fix code block width overflow causing navigation menu to shrink #472

### LINKED ISSUE #472: Ensure code-blocks take-up fixed width that's lesser than the content width
From the discussion in #463:

@mjaix says:

> I tried to understand why https://git.github.io/rev_news/2020/08/27/edition-66/, https://git.github.io/rev_news/2020/10/30/edition-68/,
 and the actual Rev News have a formatting which results in a very slim
left menu colomn. Failed :-( ... hints are appreciated. 

@jnareb says:

> If I understand it correctly, 66 and 68 edition has a very slim left menu column because the contents column is exceedingly wide due to overly long line in the code block. The menu column starts in the same place, it is just that it has less place because content column crowds it.
>
> GitRevNews-wide
> GitRevNews-narrow
>
> I wonder if it would be possible in CSS to make the _left margin_ slimmer instead of the menu column when contents column gets wide.

@sivaraam says:

> Another way to fix this would be to ensure code blocks always take up a fixed width rather than taking up as much width as it wants. I'm thinking of something similar to how code blocks in GitHub issue comments wrap withing the comment box. If that could be achieved, we wouldn't have to worry about how long the code block is.

The "adjusting left-margin" suggestion mentioned by @jnareb has been handled in #471. We still need to look into ensuring the code blocks only fixed width that's lesser than the content width. Particularly, this would ensure the rev-news edition renders well (as a single-page) on mobile screens with smaller width.

DhanashreePetare:
## Summary
Fixes #472 - Code blocks with very long lines now scroll horizontally instead of expanding the entire page width, which was causing the navigation menu to become excessively narrow.

## Problem
Code blocks containing long lines (e.g., lengthy git commands or URLs) were expanding beyond the content area, causing:
- Navigation menu to be squeezed into a very narrow column
- Poor rendering on mobile screens
- Horizontal scrolling required for the entire page

This was particularly noticeable in Rev News editions 66 and 68.

## Solution
Modified CSS for `pre` and `pre.sh_sourceCode` elements to:
- Changed `white-space: pre-wrap` to `white-space: pre` (prevents line wrapping)
- Added `overflow-x: auto` (enables horizontal scrolling within code blocks only)
- Added `max-width: 100%` (constrains code blocks to content width)

## Key Improvements Over Previous Attempt (#668)
- ✅ No font size changes - readability maintained across all screen sizes
- ✅ Only affects code blocks - no impact on other page elements
- ✅ Minimal, focused CSS changes (6 lines added/modified)
- ✅ Works seamlessly on desktop and mobile devices

## Files Changed
- `css/main.css` - Updated code block styling (7 lines modified)

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -815,13 +815,15 @@ pre {
   -moz-border-radius: 3px;
   -ms-border-radius: 3px;
   -o-border-radius: 3px;
-  white-space: pre-wrap;
+  white-space: pre;
   border-radius: 3px;
   display: block;
   padding: 10px 15px 13px 15px;
   background-color: #fff;
   border: solid 1px #efeee6;
   line-height: 18px;
+  overflow-x: auto;
+  max-width: 100%;
 }
 
 
@@ -1376,6 +1378,9 @@ pre.sh_sourceCode {
   font-size: 13px;
   font-family: Courier, monospace !important;
   line-height: 18px;
+  white-space: pre;
+  overflow-x: auto;
+  max-width: 100%;
 }
 
 


mjaix:
Hi @DhanashreePetare, 
looking at editions 66 and 68 both with Firefox 146.0 on my Linux desktop and Chrome on my Android phone, I fail to see the problem you intend to fix.
Can you provide detailed hints?
Best regards, Markus

PR #824, opened 2026-02-22
Fix broken and http links

MuxaJIbI4:
1) Fix broken links
2) Replace http -> https in links for git.github.io


chriscool:
@MuxaJIbI4 and @mjaix thanks a lot for the fixes and for reviewing them respectively.

Also sorry for not seeing this PR earlier.

About the systematic error, yeah I will try to find and fix it...

PR #827, opened 2026-03-02
Some tweaks to ideas and microprojects

sivaraam:
Based on suggestions of [Phillip Wood](https://lore.kernel.org/git/CAP8UFD2Q7gctwzGOe+rbgdXZSbDbV0dmM-cx4qt_d8nKi88=HA@mail.gmail.com/T/#t) and [@gitster](https://lore.kernel.org/all/xmqqtsv3uoc4.fsf@gitster.g/).

### DIFF
--- a/SoC-2026-Microprojects.md
+++ b/SoC-2026-Microprojects.md
@@ -73,15 +73,6 @@ It is easy though to find examples of how this can be done by
 searching the code base and the mailing list archive, as this has
 already been done for a number of languages.
 
-### Replace a run_command*() call by direct calls to C functions
-
-See for example what Junio did in
-[ffcb4e94d3](https://github.com/git/git/commit/ffcb4e94d3) (bisect: do
-not run show-branch just to show the current commit, 2021-07-27).
-
-If you can't find one please tell us, along with the command you used
-to search, so that we can remove this microproject idea.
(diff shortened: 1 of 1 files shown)


SoC-2026-Ideas.md
```diff
@@ -33,28 +33,35 @@ the same.
...
+
+The project involves careful refactoring of Git's core environment handling code,
+requiring strong C programming skills and attention to detail. Design discussions
+on the mailing list to find the best way to refactor individual variables will
+likely happen.
+
+The contributor will identify global variables that can be moved to local scope,
+implement the necessary structural changes, and ensure all affected code paths
+continue to work correctly. This includes updating tests, fixing any regressions,
+and documenting the architectural changes.
```
sivaraam: I wonder if we could something like the following here but it may be a bit more complicated, I guess.

> One promising direction is making use of functions that parse config settings gently (returning an error rather than dying), which may allow certain eagerly parsed settings to be converted to lazy parsing — but only in code paths already set up to propagate errors. This requires careful, case-by-case analysis. Refer [Stolee's work ](https://lore.kernel.org/pull.2044.v3.git.1771849615.gitgitgadget@gmail.com) in this area.

PR #839, opened 2026-03-30
rn-133: change command mentions to man page references

LemmingAvalanche:
We don’t need to use verbatim markup for commands; we can use man
page (section 1) references instead. And those that are referenced
in such a way already don’t need this markup.

Only change the style in “Discussions” in order to not infringe on the
style chosen for the “Developer Spotlight” section.

Also don’t change anything that is inside quotations (quote block/inline).

### DIFF
--- a/rev_news/drafts/edition-133.md
+++ b/rev_news/drafts/edition-133.md
@@ -42,13 +42,13 @@ surprising warning that had just circulated on Mastodon:
 
 The incident had originated in the i3 window manager project, where a
 commit message contained an unindented diff for illustration purposes.
-When Debian packagers later applied the patch using `patch(1)`, the diff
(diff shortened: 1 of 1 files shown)


rev_news/drafts/edition-133.md
```diff
@@ -93,54 +93,54 @@ that he was unsure it constituted a security attack vector, since
 someone should be reading the commit message before applying. But
 Matthias pushed back: the whole point was that nobody realized the
 behavior was there. He called it "sheer luck" that it was only a
-`sleep(1)` and not something more malicious crafted as a diff in the
+sleep(1) and not something more malicious crafted as a diff in the
```
chriscool: Back-ticks seems correct to me for sleep() here too.

PR #845, opened 2026-06-02
Minor copy edit and public-inbox upstream link correction

stepnem:


mjaix (approved):
Hello @stepnem, thanks for your corrections. They look good to me except two places.
First, I personally like Christian's "sloppy apposition" of Taylor's involvement - there I would like to leave the final decsision to Christian.
Second, I would like to revert the AKA back to aka, since the minuscules are IMHO much more common in this place. However, feel free to contradict! 
In sum, I  would like to merge your MR "as is" to be able to seamlessly proofread on top of your enhancements, since I don't want to stumble across the same details.
Best regards, Markus

stepnem:
Thank you for the review @mjaix!

By "sloppy apposition" you mean the 'apologetic about the
"trouble here"' I removed?  In the pertinent message[1]
Taylor shortly expressed his regret with "sorry for the
trouble here", the bulk of the message being the actual
solution plan.  Based on that, I found the "apologetic"
wording a bit over the top and more distracting than useful
(plus there's Taylor's "sorry again for the mess here" later
already), but I don't feel strongly about it and will gladly
defer to your (and/or Christian's) judgment.

I guess I feel a bit more strongly about AKA being more
correct/helpful/readable, and indeed, in my experience also
more common, than aka (acronyms are generally all caps), but
here, too, it shouldn't affect comprehension either way, so,
feel free to change it back should you so prefer.

Thanks again.

[1] ***@***.***/

stepnem:
after [1] in my previous commentary there should have been https://lore.kernel.org/git/aeFF5s5m76Yz7F%2FQ@nand.local/
(GH mail parsing error?)

chriscool:
Thanks @stepnem for your fixes and @mjaix for reviewing.

I also took a look, and I am fine with the current version after merging this.

mjaix:
BTW, it is fun to have you in the team.

mjaix:
I actually used "aka" in my tiny fixes just supplied, but as usual, feel free to scrutinize/improve/revert/... :-).

PR #855, opened 2026-07-29
rn-137: Add learngit.io updates to tools & sites

JackLot:
Add resource to "Git tools & sites". Minor formatting fix (remove blank newline).


chriscool:
@JackLot , sorry it looks like we missed this.

@jnareb, could you take a look for the next edition?

JackLot:
Thanks for noticing Chris. If this could be considered for the next addition, that would be appreciated. Let me know if I can do anything to help.

PR #858, opened 2026-08-26
Rev News edition 138: add crux to Other News

Sal-ami:
Adds crux to the Other News, Git tools and sites section of the edition 138 draft.

crux is a CLI that finds the exact commit which changed a program's behavior, then minimizes that commit down to the lines that caused the change. It covers cases git bisect does not: behaviors that are not pass/fail, failures that need two commits together, and regressions from dependency updates.

Source: https://github.com/Emran-goat/crux

### DIFF
--- a/rev_news/drafts/edition-138.md
+++ b/rev_news/drafts/edition-138.md
@@ -46,6 +46,8 @@ __Easy watching__
 
 __Git tools and sites__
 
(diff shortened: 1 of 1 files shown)


chriscool:
@Emran-goat, thanks for the patch. @jnareb could you take a look?

chriscool:
Sorry, it looks like we missed this one for edition 138.

Maybe we can add it to edition 139? @jnareb?

PR #859, opened 2026-09-02
rn-138: add WebTerm Learn to Git tools and sites

DaiAoki:
Adds WebTerm Learn to the "Git tools and sites" list in edition 138, with a one-line mention of its sibling sandbox webterm.app.

WebTerm Learn is a set of free Git courses that run in a simulated terminal in the browser; exercises are checked against the repository state rather than the command string. I built both, so please edit the wording freely or drop the entry if it does not fit the edition.


DaiAoki:
Superseded by #867, which adds the same entry to the edition 139 draft (the edition 138 draft this PR targeted has already been published). Closing this one.

chriscool:
@DaiAoki thanks and sorry we missed this one.

PR #865, opened 2026-09-18
Reduce Prism.js bundle size by stripping unused languages and plugins

### LINKED ISSUE #862: Some more enhancements to Prism based syntax highlighting
Right now the bundled Prism file to support syntax highlighting (cf. #490) supports these: markup (covering html, xml, svg, mathml, ssml, atom, rss), CSS, C-like, JavaScript (js), C, C++ (cpp), Bash (sh, shell), Diff, Git, Go, JSON (webmanifest), Makefile, Perl, Python (py), Ruby (rb), and YAML (yml).

The following plugins are also bundled Autoloader and File Highlight. 

I think we might be able to strip the supported languages significantly to only those that are used in the repository. This would reduce the bundle download size. 

A few other things to check:
- [ ] It seems the Prism header from the artifact downloaded from the website has been removed. I think we can retain that.
- [ ] We might need to check if it is possible to achieve the same syntax highlight in the e-mail version of the newsletter @mjaix 

cc @Amrit-raj50

Amrit-raj50:
Optimizes syntax highlighting by updating `script/prism.js` to only include languages actually used across the repository, reducing the bundle size .

### Changes
Kept languages:Markup, C-like, Bash (covering `sh`/`shell`), C, Diff, Git, and INI.
unused languages:Go, Python, Ruby, Perl, Makefile, JSON, YAML, and C++.
unused plugins:Autoloader and File Highlight.
Retained header comment:Preserved the official download configuration URL at the top of the file for future maintainability.

### Note on Email Newsletter
Regarding syntax highlighting in the email version: Prism.js cannot run in email clients because JavaScript (`<script>` tags) is blocked for security. Achieving syntax highlighting in emails would require server-side pre-rendering with inline CSS (e.g., via Rouge/Jekyll) before publishing to Mailchimp.

Closes #862


chriscool:
Thanks @Amrit-raj50 !

@sivaraam, since you were involved in the previous syntax highlighting improvements, can you take a look at this too?

Gemfile
```diff
@@ -1,8 +1,10 @@
...
-gem 'jekyll-redirect-from'
\ No newline at end of file
+gem 'jekyll', '~> 4.3'
+gem 'jekyll-redirect-from'
+gem 'webrick'
+gem 'kramdown-parser-gfm'
+gem 'csv'
+gem 'base64'
+gem 'bigdecimal'
+gem 'logger'
```
sivaraam: Why are we including additional gems?

_plugins/ruby3_compat.rb
```diff
@@ -0,0 +1,18 @@
+# Compatibility fix for Ruby 3.2+ and Ruby 4.0+
```
sivaraam: Could you explain why this is necessary in the commit message?

.bundle/config
```diff
@@ -0,0 +1,3 @@
+---
+BUNDLE_FULL_INDEX: "false"
```
sivaraam: Could you explain why this is necessary in the commit message?

sivaraam (changes requested):

Gemfile
Amrit-raj50: causing local  " bundle exec jekyll serve " to fail with LoadError without them,
but this file is pushed by mistake .

.bundle/config
Amrit-raj50: no it is not necessary in the commit mesage , i will revert all these extra 3 files

_plugins/ruby3_compat.rb
Amrit-raj50: no it is not necessary in the commit mesage , i will revert all these extra 3 files

Amrit-raj50:
@chriscool , @sivaraam could you review this , so that i can apply changes if any .

sivaraam:
Why is the prism.css unchaged? Would it not change too?

sivaraam:
Regarding the non-language content in [your comment](https://github.com/git/git.github.io/issues/862#issuecomment-5695499208):

> non-language contents --> Git Commit Graph/Log/Patch Header,Plain Text / Terminal Output / Benchmark

I think we can use the [command-line plugin](https://prismjs.com/plugins/command-line/) for terminal output.

Some helpful doc about this in README would be useful.

Amrit-raj50:
thanks @sivaraam , i will check the README

